PR #114, opened 2014-08-01
Performance data

pnispel:
Still will need to merge with master. PR for visibility


wlach:
We've chosen a bit of a different direction to go in for the perf stuff, I think it's unlikely we'll be able to use any of this as-is. :( Closing this so as to not confuse people.

PR #248, opened 2014-10-24
Bug 1062827 - Change job result names for more clarity

tojonmz:
This is the UI side change to fix Bugzilla bug [1062827](https://bugzilla.mozilla.org/show_bug.cgi?id=1062827).

In it we've agreed on these three result name changes:

* success
* testfailed -> **warnings**
* busted -> **failed**
* skipped
* exception -> **infra-failure**
* retry
* usercancel

For ease of grokking I still wonder if 'intrastructure-failure' might fit in the UI?

There is parallel work still to be done in treeherder-service and DB changes required for this commit to be landed. If I can get a local service running, I may investigate the server side. Otherwise one of the folks with full powers and access may need to do that part :)

But hopefully this saves a few minutes for you guys poking around on the UI side.

Running in its current state with no back end changes shuts off all the filter counts, to me that is expected since there are no matches.

(not fully) tested on Windows:
FF Release **33.0**
Chrome Latest Release **38.0.2125.104 m**

Adding @edmorley and @maurodoglio for review and general expertise.

### DIFF
--- a/webapp/app/css/treeherder.css
+++ b/webapp/app/css/treeherder.css
@@ -856,9 +856,9 @@ ul.failure-summary-list li .btn-xs {
 }
 
 .result-status-shading-success {background-color: rgba(2, 131, 44, 0.24);}
-.result-status-shading-testfailed {background-color: rgba(221, 102, 2, 0.25);}
-.result-status-shading-busted {background-color: rgba(144, 0, 0, 0.25);}
-.result-status-shading-exception {background-color: rgba(61, 2, 85, 0.25);}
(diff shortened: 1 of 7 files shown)


tojonmz:
Just to note here, this shouldn't be merged until service side changes have also occurred, which will be overseen by @edmorley, @maurodoglio or similar folks.

tojonmz:
Closing for now, per the Bug [thread](https://bugzilla.mozilla.org/show_bug.cgi?id=1062827#c6). We can revisit this PR later if/as needed.

PR #320, opened 2014-12-30
Bug 1110552 - faster ``showHideJob`` perf

camd:
Switched from using jQuery for showing/hiding and finding
classes to using DOM calls.  This ended up being about 17% faster when
toggling unclassified failures.

**Update**
With some suggestions from Mauro and the use of ``.classList.add()`` this is now benchmarking at 30% faster to show/add a job.

### DIFF
--- a/webapp/app/js/directives/clonejobs.js
+++ b/webapp/app/js/directives/clonejobs.js
@@ -569,33 +569,33 @@ treeherder.directive('thCloneJobs', [
         // Note: I was using
         //     jobEl.style.display = "inline";
         //     jobEl.className += " filter-shown";
-        // but that didn't work reliably with the jquery selectors
+        // but the classname setting didn't work reliably with the jquery selectors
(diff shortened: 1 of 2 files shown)


webapp/app/js/directives/clonejobs.js
```diff
@@ -569,33 +569,33 @@ treeherder.directive('thCloneJobs', [
...
+        // but the classname setting didn't work reliably with the jquery selectors
         // when it came to hiding/showing platforms and groups.  Jquery
         // couldn't detect that I'd added or removed ``filter-shown`` in
         // all cases.  So, while this is a bit slower, it's reliable.
+        //
+        // It would be great to be able to do this without adding/removing a class
         if (show) {
-            job.css("display", "inline-block");
-            job.addClass("filter-shown");
+            job[0].classList.add("filter-shown");
```
camd: adding the class this way was WAY faster than use ``.addClass``.  I test it on Chrome and Safari and it works great.

PR #324, opened 2015-01-03
Bug 1111157 - Give visited links their own color

ehsan:


tojonmz:
Hi @ehsan I tried your branch out of interest. The only major item I observed not receiving your visited link color are all Failure Summary suggested bugs. Was yours working running your local client?

visitedfailuresummarylinks

camd:
@ehsan - thanks for tackling this.  Though I agree with Ed.  I wouldn't want to modify ``bootstrap.css`` directly.  Can you add this as an override in ``treeherder.css``?  That way, when we upgrade to the latest bootstrap at some point, we don't have to know where to modify their ``.css`` file.

Thanks!

edmorley:
The failure summary box css is at:
https://github.com/mozilla/treeherder-ui/blob/master/webapp/app/css/treeherder.css#L278

@ehsan if it helps, the unminified css is available on treeherder-dev.allizom.org

camd:
Looks great!  Thanks!

PR #334, opened 2015-01-10
Bug 1090531 - fix refresh on "push not processed yet" page

camd:
This fixes bug 1090531 as well as bug Bug 1071002

The code for polling was inaccurate in several ways.  
* The polling wasn't using the params on the address bar at all.  So I added that.  
* And in some cases, it wasn't doing the polling when it should (it should still poll with ``startdate`` and ``fromchange`` set).

### DIFF
--- a/webapp/app/js/models/resultsets.js
+++ b/webapp/app/js/models/resultsets.js
@@ -33,18 +33,15 @@ treeherder.factory('ThResultSetModel', [
     // the primary data model
     var repositories = {};
 
-    var updateQueueInterval = 10000;
-
     var resultSetPollers = {};
     var resultSetPollInterval = 60000;
     var jobPollInterval = 90000;
     var pollDelayMin = 1000;
     var pollDelayMax = 60000;
 
-    // if any of these params are in the url, don't poll for new data
-    var noPollingParameters = [
-        'fromchange', 'tochange', 'startdate', 'enddate', 'revision'
-    ];
+    // Keys that, if present on the url, must be passed into the resultset
+    // polling endpoint
+    var rsPollingKeys = ['tochange', 'enddate', 'revision', 'author'];
 
     // changes to the url for any of these fields should reload the page
     // because it changes the query to the db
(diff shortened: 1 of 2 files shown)


tojonmz:
@camd cool! I will see if I can get try access to test this fix also. I noticed a couple trivial improvements (maybe :) ) in the patch. Perhaps I'll comment and please feel free to ignore as applicable :)

webapp/app/js/models/resultsets.js
```diff
@@ -41,9 +41,10 @@ treeherder.factory('ThResultSetModel', [
...
     var pollDelayMax = 60000;
 
-    // if any of these params are in the url, don't poll for new data
-    var noPollingParameters = [
-        'fromchange', 'tochange', 'startdate', 'enddate', 'revision'
+    // Keys that, if present on the url, must be passed into the resultset
+    // polling endpoint
+    var rsPollingKeys = [
+        'tochange', 'enddate', 'revision', 'author'
     ];
```
tojonmz: I wonder if this var can exist on one line since it's now under 80 chars.

webapp/app/js/models/resultsets.js
```diff
@@ -61,30 +62,52 @@ treeherder.factory('ThResultSetModel', [
...
-                if( (repositories[$rootScope.repoName].resultSets.length > 0) &&
-                    (repositories[$rootScope.repoName].loadingStatus.prepending === false) ){
+        // Register resultset poller if it's not registered
+        $interval(function(){
 
+            // This case is if we already have at least 1 resultset and we're
+            // polling for more.
+            // If we already have one, and the revision param is passed in,
+            // don't poll for more, because "there can be only one."
+            if( (repositories[$rootScope.repoName].resultSets.length > 0) &&
```
tojonmz: Was just curious about the non-space after the `if` and the white space between the `( ( `, similarly on the `else if(` on line 88. For my own interest mostly :) ...is it a particular editor convention on long conditions?

webapp/app/js/models/resultsets.js
```diff
@@ -61,30 +62,52 @@ treeherder.factory('ThResultSetModel', [
...
 
-                } else if( (repositories[$rootScope.repoName].resultSets.length === 0) &&
-                           (repositories[$rootScope.repoName].loadingStatus.prepending === false) ){
+            } else if( (repositories[$rootScope.repoName].resultSets.length === 0) &&
+                       (repositories[$rootScope.repoName].loadingStatus.prepending === false) ){
 
-                    fetchResultSets(
-                        $rootScope.repoName, defaultResultSetCount);
+                fetchResultSets(
+                    $rootScope.repoName, defaultResultSetCount);
```
tojonmz: Was wondering about this perhaps fitting on one line also.

webapp/app/js/models/resultsets.js
camd: That was actually Jeads' style.  And I only changed the indentation on that code, so I didn't change it.  though I think it's worth changing that.  I prefer doing it the way I think you're inferring.  I'll change it.

webapp/app/js/models/resultsets.js
camd: this is one I can often go either way on.  But, yeah.  Probably defaulting to a single line where possible is better.  :)

webapp/app/js/models/resultsets.js
```diff
@@ -59,32 +58,56 @@ treeherder.factory('ThResultSetModel', [
...
-                if( (repositories[$rootScope.repoName].resultSets.length > 0) &&
-                    (repositories[$rootScope.repoName].loadingStatus.prepending === false) ){
+            // This case is if we already have at least 1 resultset and we're
+            // polling for more.
+            // If we already have one, and the revision param is passed in,
+            // don't poll for more, because "there can be only one."
 
+            if((rsData.resultSets.length > 0) &&
+               (!rsData.loadingStatus.prepending)) {
+                if (doResultSetPolling(rsPollingParams)) {
```
camd: if this comes out false in ``doResultSetPolling`` then we should cancel the interval.

webapp/app/js/models/resultsets.js
```diff
@@ -59,32 +58,56 @@ treeherder.factory('ThResultSetModel', [
         'exclusion_state'
     ];
 
-    var registerResultSetPollers = function(){
+    var registerResultSetPollers = function() {
```
camd: Ed: I was wrong, the resultsets are not loaded before we first call this function.  This function performs the first load of resultsets on page load.  So I can't check ``doResultSetPolling`` here.   It would be premature.  But I will still cancel when it calls that method and gets a ``false`` after initial data load.

camd:
@edmorley : ok, I think these changes address what we spoke of today and will probably introduce whirled peas.  Or at least make philor crack a small smile....

PR #336, opened 2015-01-13
Bug 1077554 - Use target="_blank" in fewer places

edmorley:
It's currently (a) incorrectly being used for internal navigation links,
and (b) being used in cases where the user would likely not expect it.
eg: the standalone help page, where having to use the back button would
not lose any state (unlike the main Treeherder UI).

### DIFF
--- a/webapp/app/logviewer.html
+++ b/webapp/app/logviewer.html
@@ -20,8 +20,7 @@
                             <td ng-if="label == 'revision'" class="break-word">
                                 {{value}}
                                 <a href="{{logRevisionFilterUrl}}"
-                                   title="open resultset in new tab"
-                                   target="_blank"
+                                   title="open resultset"
                                    class="repo-link"
                                    ng-cloak>link</a>
                             </td>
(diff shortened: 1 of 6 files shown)


webapp/app/logviewer.html
```diff
@@ -21,7 +21,6 @@
                                 {{value}}
                                 <a href="{{logRevisionFilterUrl}}"
                                    title="open resultset in new tab"
```
camd: we should update this title text, too.

webapp/app/partials/main/jobs.html
```diff
@@ -13,8 +13,7 @@
         <span class="btn btn-default btn-sm glyphicon glyphicon-download" ng-show="isLoadingJobs"></span>
         <span>
           <a href="{{::revisionResultsetFilterUrl}}"
-             title="open this resultset in new tab"
-             target="_blank">{{::resultsetDateStr}}</a> - </span>
+             title="open this resultset in new tab">{{::resultsetDateStr}}</a> - </span>
```
camd: update the title text here as well.  not opening in a new tab anymore.  :)

PR #339, opened 2015-01-14
bug 1071541 - remove local.conf.js from the grunt build

camd:
Now that we default to the same host for the service, we don’t need a
local.conf.js file to work properly.  Removing this prevents the
accident of stage pointing to prod or vice-versa, etc.


tojonmz:
Does any documentation need to be updated for local development setup (eg. without vagrant). Or is this a seamless change.

camd:
This won't affect local or vagrant setups.  It's only ones that use the grunt build like stage and prod.

PR #342, opened 2015-01-19
When selecting a Talos run, for each uploaded profile zip, show a link for opening it directly in Cleopatra.

mstange:
This will work once these are fixed: https://github.com/bgirard/cleopatra/pull/54 , https://bugzilla.mozilla.org/show_bug.cgi?id=1121571 , https://bugzilla.mozilla.org/show_bug.cgi?id=1122848

Here's the motivation, copied from https://bugzilla.mozilla.org/show_bug.cgi?id=1122848#c0 :

I'd like to make Talos profiles accessible in cleopatra (our profile visualization / consumption tool) through a simple link on treeherder.

When you push to try with "mozharness: --spsProfile" at the end of the try chooser commit message, Talos will run its performance tests under a profiler, create zip files that contain the collected profiles, and put them in the MOZ_UPLOAD_DIR. These files can then be downloaded using the "artifact uploaded" links on treeherder. At the moment, the user then has to unzip those files, load http://people.mozilla.com/~bgirard/cleopatra/ in a browser, and drag one of the unzipped profiles into the file chooser on cleopatra.

I'd like to simplify this process by having treeherder insert a link to 
http://people.mozilla.com/~bgirard/cleopatra/?zippedProfile=[url to the zip].
Then cleopatra would grab the zip, extract it right in the browser, ask the user to select one of the contained profiles, and then display that profile.

### DIFF
--- a/webapp/app/plugins/talos/main.html
+++ b/webapp/app/plugins/talos/main.html
@@ -16,3 +16,10 @@
     </li>
   </ul>
 </span>
+
+<ul>
+  <li ng-repeat="line in job_details"
+      ng-if="line.content_type == 'link' && line.value.substr(-8) == '.sps.zip'">
+    <a href="http://people.mozilla.com/~bgirard/cleopatra/?zippedProfile={{line.url}}">Open {{line.value}} in Cleopatra</a>
+  </li>
+</ul>


mstange:
Here's an example try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=ec8bf3a470d1

edmorley:
I think TinderboxPrints of type url are a better solution for this - it's more scalable than adding another conditional for each additional piece of metadata, and other consumers of the API can get them too. We'll be highlighting TinderboxPrints more in the UI in the future too, plus giving it a better name and more structured format.

edmorley:
I think whatever the solution we take, this should be done upstream of the UI - whether that be custom handling during ingestion, or else in Talos itself (by generating the full TinderboxPrint URL).

mstange:
Doing it in Talos itself is a little tricky - the TinderboxPrint is done during blob upload, after Talos has finished. (Look for "(blobuploader) - INFO - TinderboxPrint:" in http://ftp.mozilla.org/pub/mozilla.org/firefox/try-builds/mstange@themasta.com-ec8bf3a470d1/try-win32/try_win7-ix_test-g1-bm109-tests1-windows-build348.txt.gz ). Talos doesn't know the final URL before that, so it can't do the TinderboxPrint. And the interface between Talos and blob uploading is basically just "drop a file in this directory" - we can't pass any data along with it. That's why I've resorted to the ".sps.zip" file extension.

Where is "custom handling during ingestion" implemented? Should I go that route?

mstange:
All the other pieces of the puzzle have now landed. Can you give me some guidance here, please?

wlach:
Maybe I'm missing something here, but I feel the solution proposed here by @mstange is pretty much fine. We're already parsing out the value via TinderboxPrint, we just need some custom logic to display it in a way that it's easily linkable. That needs to live somewhere, treeherder-ui seems like a fine place for it, certainly better than Talos.

edmorley:
Let's land this for now, and longer term if other such use-cases come up, see if we can implement a more generic solution in treeherder-service to sort through the list of blobber files and add them as artefacts with set types ('manifest', 'machine-readable-log', 'talos-profile' etc).

mstange:
Thanks!

PR #343, opened 2015-01-20
Bug 1120133 - Enhance copy of logviewer/raw log icon links

tojonmz:
This work fixes Bugzilla bug [1120133](https://bugzilla.mozilla.org/show_bug.cgi?id=1120133).

This allows the user to hover over the logviewer or raw log icon, which injects into a placeholder clipboard div a specified value. This can then be directly copied to the users' system clipboard.

In the change below, we mimic the Trello solution in angular. Here's the general workflow scenario:

* User hovers over a log icon

...mouseenterhover2

* User simply issues a copy to their clipboard via **ctrl+c** or **cmd+c**

* User pastes it somewhere, for example in channel

...linkpaste

So the previous workflows, are no longer required.

Everything seems to be working generally fine, at least on windows. If someone can try it with vagrant which uses the intended $scope.lvFullUrl link fragment value of `/ui/` that would be great. I was using a local value of `/app/` during local development.

If someone feels strongly we should include this in Help or expand the tooltip title for logviewer/raw log we can, but it seemed a bit excessive, since this is quite a power user type of feature and clutters the tooltip.

Tested on Windows:
FF Release **35.0**
Chrome Latest Release **39.0.2171.99 m**

### DIFF
--- a/webapp/app/css/treeherder.css
+++ b/webapp/app/css/treeherder.css
@@ -28,6 +28,23 @@ body {
     visibility: hidden;
 }
 
+#clipboard-container {
+    position: fixed;
+    top: 0px;
+    left: 0px;
+    width: 0px;
+    height: 0px;
+    z-index: 100;
+    display: none;
+    opacity: 0;
+}
+
+#clipboard {
+    width: 1px;
+    height: 1px;
+    padding: 0px;
+}
+
 .th-navbar {
     background-color: #273038;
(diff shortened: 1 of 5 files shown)


tojonmz:
On discussion with @camd in channel on tooltip vs. help, we decided on adding a help entry for the behaviour. We felt the tooltip would get too long.

helpaddition

PR #346, opened 2015-01-22
Bug 1124792 - Failure summary: Add bug resolution to closed bug tooltips

edmorley:
This:
* Removes unnecessary resolution check for open_recent bugs (since the bugs by definition are always open).
* If a bug is resolved/closed, shows the resolution (eg 'FIXED', 'INVALID') in the tooltip for the bug link. With this change, there would have been three places that were conditional on |bug.resolution != ''| so I've switched to an ng-if and two versions of the link.


edmorley:
screenshot

PR #347, opened 2015-01-23
Bug 1125126 - Make the visited link color less hideous

KWierso:
I think the plain purple is less terrible than what's on there currently.

### DIFF
--- a/webapp/app/css/treeherder.css
+++ b/webapp/app/css/treeherder.css
@@ -12,7 +12,7 @@ body {
 }
 
 a:visited {
-    color: #a842ca;
+    color: #purple;
 }
 
 #global-container {
@@ -731,7 +731,7 @@ ul.failure-summary-list li .btn-xs {
     color: #0000ee;
 }
 .failure-summary-bugs a:visited {
-    color: #a842ca;
+    color: #purple;
 }
 
 .failure-summary-bugs button {


KWierso:
This also should hopefully not apply to the Show/Hide more link, since that shouldn't count as a visited link.

webapp/app/css/treeherder.css
```diff
@@ -11,8 +11,8 @@ body {
     line-height: inherit; /* override bootstrap */
 }
 
-a:visited {
-    color: #a842ca;
+a:visited:not(.show-hide-more) {
```
edmorley: I think I'd prefer avoiding the browser seeing show/hide more as a visited link in the first place - eg by replacing show/hide more with a span (in another PR, don't worry about it here), since <a> is not really appropriate, since it's not really a link. ie: http://stackoverflow.com/a/135248

@camd - thoughts?

Thank you for the PR though - much better colour :-)

KWierso: Squashed in a new commit that removes the show/hide selector. This PR now only changes visited link colors.

camd:
@edmorley : I agree.  show/hide more shouldn't be seen as a "visited" style link.

edmorley:
Thanks :-)

PR #348, opened 2015-01-24
Bug 1120133 - Remove ui fragment from logviewer hover link

tojonmz:
Supplemental fix to Bugzilla bug [1120133](https://bugzilla.mozilla.org/show_bug.cgi?id=1120133).

This removes the `/ui` path fragment from the logviewer icon 'hover' url, as it is not needed on production.

Someone with vagrant will want to sanity-check this change locally before merge. I have checked the existing production environment works without the fragment, so it should work.

Not locally tested, per above.

Adding @camd for review and a local check.

### DIFF
--- a/webapp/app/plugins/controller.js
+++ b/webapp/app/plugins/controller.js
@@ -148,7 +148,7 @@ treeherder.controller('PluginCtrl', [
                         });
                     }
                     $scope.lvUrl = thUrl.getLogViewerUrl($scope.job.id);
-                    $scope.lvFullUrl = location.origin + "/ui/" + $scope.lvUrl;
+                    $scope.lvFullUrl = location.origin + "/" + $scope.lvUrl;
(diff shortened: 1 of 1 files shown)


webapp/app/plugins/controller.js
```diff
@@ -148,7 +148,7 @@ treeherder.controller('PluginCtrl', [
                         });
                     }
                     $scope.lvUrl = thUrl.getLogViewerUrl($scope.job.id);
-                    $scope.lvFullUrl = location.origin + "/ui/" + $scope.lvUrl;
+                    $scope.lvFullUrl = location.origin + $scope.lvUrl;
```
edmorley: This needs an additional '/' before the log URL I think? :-)

tojonmz: Ya, agreed. Will change and re-push.

PR #349, opened 2015-01-25
Fix bug 1108837 - "Click to add a related bug" text should probably also...

kewisch:
... be clickable.

This also makes a few more small enhancements to the bug field, I think they are too minor to open a new bug for each.


camd:
@kewisch  -  Thanks for the patch!  Though I'm not able to actually enter a bug.  It always asks me to "please enter a bug id".  I'm on OSX, for context.

The click-ableness and escape to close works great, however.  :)  Would you see if you can reproduce?

camd:
this looks and works great!  Thanks for the fix!  :)

PR #351, opened 2015-01-29
[Fixes #1081041] Linkify PR and bugs in commit messages

mishravikas:
Some example commit messages which do not follow the pattern "Bug XXXX" and match with this pr:
https://treeherder.mozilla.org/#/jobs?repo=gaia-try&revision=0551b4faf451
https://treeherder.mozilla.org/#/jobs?repo=gaia-try&revision=0f6f9e5ce777
https://treeherder.mozilla.org/#/jobs?repo=gaia-try&revision=e1e278656843

### DIFF
--- a/webapp/app/js/filters.js
+++ b/webapp/app/js/filters.js
@@ -39,11 +39,21 @@ treeherder.filter('stripHtml', function() {
 
 treeherder.filter('linkifyBugs', function() {
     return function(input) {
-        var re = new RegExp('(?:Bug (\\d+))', 'ig');
         var str = input || '';
-        return str.replace(re,
-            '<a href="https://bugzilla.mozilla.org/show_bug.cgi?id=$1" target="_blank" title="Bug $1">$1</a>'
-        );
+        var bug_matches = /.*-- ([0-9]+)|.*bug-([0-9]+)|.*Bug ([0-9]+)/ig.exec(str);
+        var pr_matches = /PR#([0-9]+)/i.exec(str);
+        if (pr_matches) {
+            var pr_url = "https://github.com/mozilla-b2g/gaia/pull/" + pr_matches[1];
+            var pr_hyperlink = '<a href="' + pr_url + '">' + pr_matches[1] + '</a>';
+            str = str.replace(pr_matches[1], pr_hyperlink);
+        }
+        if (bug_matches) {
+            var bug_match = bug_matches[1] || bug_matches[2] || bug_matches[3];
+            var bug_url = "https://bugzilla.mozilla.org/show_bug.cgi?id=" + bug_match;
+            var bug_hyperlink = '<a href="' + bug_url + '">' + bug_match + '</a>';
+            str = str.replace(bug_match, bug_hyperlink);
+        }
+        return str;
     };
 });
 


webapp/app/js/filters.js
```diff
@@ -39,11 +39,21 @@ treeherder.filter('stripHtml', function() {
 
 treeherder.filter('linkifyBugs', function() {
     return function(input) {
-        var re = new RegExp('(?:Bug (\\d+))', 'ig');
-        var str = input || '';
-        return str.replace(re,
-            '<a href="https://bugzilla.mozilla.org/show_bug.cgi?id=$1" target="_blank" title="Bug $1">$1</a>'
-        );
+        if (input) {
```
wlach: If `input` is not undefined, this function will now return undefined, whereas before it would return an empty string. Was this intentional?

mishravikas: no it wasn't intentional, I'll update the pr

webapp/app/js/filters.js
```diff
@@ -39,11 +39,21 @@ treeherder.filter('stripHtml', function() {
...
+            var re = /.*-- (bug-)?([0-9]+)/i; //regex for matching bug number
+            var bug_matches = re.exec(input);
```
wlach: Let's just do something like var bug_matches = /PR#([0-9]+)/i.exec(input) instead of defining (and then redefining) a temporary re variable.

wlach: Also, I don't think this will work for the old case where we have "Bug 12345 - Fix the foo" which is the style used in mozilla-central. If you want help crafting a regular expression that will work for that case feel free to ask in #ateam or #treeherder. :)

webapp/app/js/filters.js
```diff
@@ -39,11 +39,21 @@ treeherder.filter('stripHtml', function() {
...
+            var bug_url = "https://bugzilla.mozilla.org/show_bug.cgi?id="+bug_matches[2];
+            var re = /PR#([0-9]+)/i; //regex for matching PR number
+            var pr_matches = re.exec(input);
+            var pr_url = "https://github.com/mozilla-b2g/gaia/pull/" + pr_matches[1];
+            var str = input || '';
+            var pr_hyperlink ='<a href="'+pr_url+'">'+pr_matches[1]+'</a>';
+            str = str.replace(pr_matches[1],pr_hyperlink);
+            var bug_hyperlink ='<a href="'+bug_url+'">'+bug_matches[2]+'</a>';
+            str = str.replace(bug_matches[2],bug_hyperlink);
```
wlach: Do these work if bug_matches or pr_matches are null (which can happen if the regex doesn't match)?

mishravikas: If the regex does not matches for either of the bug_matches and pr_matches, should it return an empty string?

edmorley:
If it helps, the TBPL regexes for this are here https://hg.mozilla.org/webtools/tbpl/file/547a7de112ac/js/UserInterface.js#l777

webapp/app/js/filters.js
wlach: No, we should just return the string as-is in those cases. The intention is to replace a string like "Bug XXXX - Blah" with something like: "<a href="http://bugzilla..mozilla.org/show_bug.cgi?id=XXX">Bug XXXXX</a> - Blah. But if we don't match for whatever reason (e.g. someone mistyped and put in "Bg XXXX - Blah" we should maintain that)

webapp/app/js/filters.js
```diff
@@ -39,11 +39,25 @@ treeherder.filter('stripHtml', function() {
...
-        var re = new RegExp('(?:Bug (\\d+))', 'ig');
-        var str = input || '';
-        return str.replace(re,
-            '<a href="https://bugzilla.mozilla.org/show_bug.cgi?id=$1" target="_blank" title="Bug $1">$1</a>'
-        );
+        if (input) {
+            var re = /.*-- ([0-9]+)|.*bug-([0-9]+)|.*Bug ([0-9]+)/ig; //Matches 3 paterrns "<-- XXX" , "bug-XXX" and "Bug XXX"
+            var bug_matches = re.exec(input);
+            var pr_matches = /PR#([0-9]+)/i.exec(input);
+            if (pr_matches && bug_matches ) { //Check if there are any matches, if not return the 'input'
```
wlach: We should handle pr_matches and bug_matches seperately. The way this clause is working is that we'll only replace one if we replace them both. Some messages will only reference bug #'s, not PR's.

webapp/app/js/filters.js
```diff
@@ -39,11 +39,25 @@ treeherder.filter('stripHtml', function() {
...
+                var bug_match = bug_matches[1] || bug_matches[2] || bug_matches[3];
+                var bug_url = "https://bugzilla.mozilla.org/show_bug.cgi?id="+bug_match;
+                var pr_url = "https://github.com/mozilla-b2g/gaia/pull/" + pr_matches[1];
+                var str = input;
+                var pr_hyperlink ='<a href="'+pr_url+'">'+pr_matches[1]+'</a>';
+                str = str.replace(pr_matches[1],pr_hyperlink);
+                var bug_hyperlink ='<a href="'+bug_url+'">'+bug_match+'</a>';
+                str = str.replace(bug_match,bug_hyperlink);
```
wlach: Did you test this to be sure it works? Also, you should have a space between ',' and the next argument when calling a function. You should also space expressions like 'foo'+'bar' to 'foo' + 'bar'.

webapp/app/js/filters.js
wlach: Ok, the logic is good now, however I think I still prefer the earlier way this was implemented, where we created a variable str equal to '' if input was not defined, and just returned that (var str = input || ''). That removes the need for if..else.

webapp/app/js/filters.js
```diff
@@ -39,11 +39,26 @@ treeherder.filter('stripHtml', function() {
...
-        var re = new RegExp('(?:Bug (\\d+))', 'ig');
-        var str = input || '';
-        return str.replace(re,
-            '<a href="https://bugzilla.mozilla.org/show_bug.cgi?id=$1" target="_blank" title="Bug $1">$1</a>'
-        );
+        if (input) {
+            var re = /.*-- ([0-9]+)|.*bug-([0-9]+)|.*Bug ([0-9]+)/ig; //Matches 3 paterrns "<-- XXX" , "bug-XXX" and "Bug XXX"
+            var bug_matches = re.exec(input);
+            var pr_matches = /PR#([0-9]+)/i.exec(input);
+            var str = input;
```
wlach: This line can go away if you implement my earlier suggestion.

webapp/app/js/filters.js
```diff
@@ -39,11 +39,26 @@ treeherder.filter('stripHtml', function() {
 
 treeherder.filter('linkifyBugs', function() {
     return function(input) {
-        var re = new RegExp('(?:Bug (\\d+))', 'ig');
-        var str = input || '';
-        return str.replace(re,
-            '<a href="https://bugzilla.mozilla.org/show_bug.cgi?id=$1" target="_blank" title="Bug $1">$1</a>'
-        );
+        if (input) {
+            var re = /.*-- ([0-9]+)|.*bug-([0-9]+)|.*Bug ([0-9]+)/ig; //Matches 3 paterrns "<-- XXX" , "bug-XXX" and "Bug XXX"
```
wlach: Let's move the comment above (also, typo: paterrns -> patterns) and just execute that raw regex, instead of creating a temporary variable re.

webapp/app/js/filters.js
```diff
@@ -39,11 +39,26 @@ treeherder.filter('stripHtml', function() {
...
+            var pr_matches = /PR#([0-9]+)/i.exec(input);
+            var str = input;
+            if (pr_matches) {
+                var pr_url = "https://github.com/mozilla-b2g/gaia/pull/" + pr_matches[1];
+                var pr_hyperlink = '<a href="' + pr_url + '">' + pr_matches[1] + '</a>';
+                str = str.replace(pr_matches[1], pr_hyperlink);
+            }
+            if (bug_matches) {
+                var bug_match = bug_matches[1] || bug_matches[2] || bug_matches[3];
+                var bug_url = "https://bugzilla.mozilla.org/show_bug.cgi?id="+bug_match;
```
wlach: Spaces between the '+' and the expressions please.

wlach:
Merged manually.

PR #354, opened 2015-02-03
Bug 1127454 - Remove status counts

camd:
**READY FOR MERGE**

This removes the status counts and replaces them with a % progress and count of "in progress" jobs like so:
Before:
screenshot 2015-02-06 15 45 16
 
After:
screenshot 2015-02-06 15 41 37

This also attempts to "remove clutter" from the resultset status bar.

Also fixed:
* global collapse button just collapses revisions, not everything.  remove drop-down.
* cleaned up resultset buttons and moved right.
* global filtering of coalesced (was not functioning)
* Fixed counts of pending vs. coalesced that was off
* removed counting code in ``clonejobs.js`` because ``resultset_store`` already does it
* made failed jobs have transparent background when classified
* removed dead code from clonejobs for showing only revisions without jobs.

### DIFF
--- a/webapp/app/css/treeherder.css
+++ b/webapp/app/css/treeherder.css
@@ -1242,6 +1242,33 @@ fieldset[disabled] .btn-orange.active {
   border-color: #dd6602;
 }
 
+.btn-orange-classified {
+  color: #dd6602;
+}
+.btn-orange-classified:hover,
+.btn-orange-classified:focus,
+.btn-orange-classified:active,
+.btn-orange-classified.active {
+  background-color: #c45a02;
+  border-color: #aa4f02;
+  color: white;
+}
+.btn-orange-classified.disabled:hover,
+.btn-orange-classified.disabled:focus,
+.btn-orange-classified.disabled:active,
+.btn-orange-classified.disabled.active,
+.btn-orange-classified[disabled]:hover,
+.btn-orange-classified[disabled]:focus,
+.btn-orange-classified[disabled]:active,
+.btn-orange-classified[disabled].active,
+fieldset[disabled] .btn-orange-classified:hover,
+fieldset[disabled] .btn-orange-classified:focus,
+fieldset[disabled] .btn-orange-classified:active,
+fieldset[disabled] .btn-orange-classified.active {
+  background-color: #dd6602;
+  border-color: #dd6602;
+}
+
 .btn-orange-count-classified {
   background-color: white;
   border-color: #bfbfbf;
@@ -1281,6 +1308,34 @@ fieldset[disabled] .btn-red:active,
 fieldset[disabled] .btn-red.active {
   background-color: #c2020e;
   border-color: #c2020e;
+}
+
+.btn-red-classified {
+  color: #90010a;
+}
+.btn-red-classified:hover,
+.btn-red-classified:focus,
+.btn-red-classified:active,
+.btn-red-classified.active {
+  background-color: #a9020c;
+  border-color: #90010a;
+  color: white;
+}
+.btn-red-classified.disabled:hover,
+.btn-red-classified.disabled:focus,
+.btn-red-classified.disabled:active,
+.btn-red-classified.disabled.active,
+.btn-red-classified[disabled]:hover,
+.btn-red-classified[disabled]:focus,
+.btn-red-classified[disabled]:active,
+.btn-red-classified[disabled].active,
+fieldset[disabled] .btn-red-classified:hover,
+fieldset[disabled] .btn-red-classified:focus,
+fieldset[disabled] .btn-red-classified:active,
+fieldset[disabled] .btn-red-classified.active {
+  background-color: #c2020e;
+  border-color: #c2020e;
+  color: white;
 }
 
 .btn-red-count-classified {
@@ -1385,6 +1440,34 @@ fieldset[disabled] .btn-purple:hover,
 fieldset[disabled] .btn-purple:focus,
 fieldset[disabled] .btn-purple:active,
 fieldset[disabled] .btn-purple.active {
+  background-color: #9002c2;
+  border-color: #9002c2;
+  color: white;
+}
+
+.btn-purple-classified {
+  color: #6f0296;
+}
+.btn-purple-classified:hover,
+.btn-purple-classified:focus,
+.btn-purple-classified:active,
+.btn-purple-classified.active {
+  background-color: #7d02a9;
+  border-color: #6b0190;
+  color: white;
+}
+.btn-purple-classified.disabled:hover,
+.btn-purple-classified.disabled:focus,
+.btn-purple-classified.disabled:active,
+.btn-purple-classified.disabled.active,
+.btn-purple-classified[disabled]:hover,
+.btn-purple-classified[disabled]:focus,
+.btn-purple-classified[disabled]:active,
+.btn-purple-classified[disabled].active,
+fieldset[disabled] .btn-purple-classified:hover,
+fieldset[disabled] .btn-purple-classified:focus,
+fieldset[disabled] .btn-purple-classified:active,
+fieldset[disabled] .btn-purple-classified.active {
   background-color: #9002c2;
   border-color: #9002c2;
   color: white;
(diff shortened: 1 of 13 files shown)


tojonmz:
@camd looks very nice!

Maybe it would help separate the percentage-done element, if it used a different tone from the SHA? I believe the running filter-text used `#6f6d70`. Or maybe some similar light grey when in progress. Just an idea. But looks awesome.

camd:
Here's a "take-two" on the % complete part:

screenshot 2015-02-04 09 47 53

tojonmz:
Looks very nice.

webapp/app/css/treeherder.css
```diff
@@ -422,8 +422,29 @@ th-watched-repo {
...
+}
+
+.result-set-left {
+    display: flex;
+    display: -webkit-flex;
+    -webkit-flex-flow: row wrap;
     flex-flow: row wrap;
+    flex: auto;
     align-items: center;
+
```
tojonmz: Just noticed this extra line in passing :)

webapp/app/js/services/jobfilters.js
```diff
@@ -486,7 +486,8 @@ treeherder.factory('thJobFilters', [
     var _matchesDefaults = function(field, values) {
         $log.debug("_matchesDefaults", field, values);
         if (DEFAULTS.hasOwnProperty(field)) {
-            return _.intersection(DEFAULTS[field].values, values).length === DEFAULTS[field].values.length;
+            return values.length === DEFAULTS[field].values.length &&
```
camd: without this, we weren't noticing that you'd added ``coalesced`` to the global filters, so it wasn't showing them, due to the way ``_.intersection`` works.

webapp/app/js/directives/clonejobs.js
```diff
@@ -646,49 +649,6 @@ treeherder.directive('thCloneJobs', [
         }
     };
 
-    /**
-     * Reset the counts on a single resultset.  This should happen once per
-     * resultset when the platforms in a resultset have been updated
-     * (by loading initial or updated data).
-     *
-     */
-    var resetCounts = function(resultSet){
```
camd: this counting is all handled in the ``resultset_store`` from mauro's changes.  So this was wasted effort.

webapp/app/js/directives/clonejobs.js
```diff
@@ -236,7 +236,10 @@ treeherder.directive('thCloneJobs', [
             jobStatus.key = key;
             if(parseInt(job.failure_classification_id, 10) > 1){
                 jobStatus.value = job.job_type_symbol + '*';
-            }else{
+                if (jobStatus.btnClassClassified) {
+                    jobStatus.btnClass = jobStatus.btnClassClassified;
```
camd: show the "classified" version of a failed jobs with a transparent background when it's been, well... classified.

webapp/app/js/directives/resultsets.js
```diff
@@ -72,52 +72,6 @@ treeherder.directive('thResultCounts', [
     };
 }]);
 
-treeherder.directive('thResultStatusCount', [
```
camd: this used to display all the job counts for each resultset.

edmorley:
> global collapse button just collapses revisions, not everything. remove drop-down.
> cleaned up resultset buttons and moved right.
> global filtering of coalesced (was not functioning)
> Fixed counts of pending vs. coalesced that was off
> removed counting code in clonejobs.js because resultset_store already does it
> made failed jobs have transparent background when classified
> removed dead code from clonejobs for showing only revisions without jobs.

Would you mind breaking some of these parts out into another commit, to make this more readable? (just the ones that are easily separated out). Using either SourceTree or else git add -p this should only take a couple of mins :-)

For how to split an existing commit, plus how to use git add -p to selectively stage hunks, see http://stackoverflow.com/a/1440200 and http://cbx33.github.io/gitt/afterhours5-1.html

webapp/app/js/directives/clonejobs.js
edmorley: I'm presuming we need the conditional because some result states don't have the solid border to start with? Could you add a comment explaining? eg
// For result types that are displayed more prominently when unclassified, switch to the more subtle unclassified style.

webapp/app/css/treeherder.css
```diff
@@ -418,10 +418,30 @@ th-watched-repo {
...
+    -webkit-flex-flow: row wrap;
     flex-flow: row wrap;
+    flex: auto;
     align-items: center;
 }
 
+.result-set-progress {
+    color: #6f6d70;
+    font-style: italic;
+    margin-left: 10px;
```
edmorley: I think we could move this further to the right - ie on my window size the "In progress" counts appear really close the the revision list, rather than above the job counts. Perhaps we should be aligning relative to the buttons on the right hand side instead? :-)

webapp/app/js/directives/persona.js
```diff
@@ -19,6 +19,8 @@ treeherder.directive('personaButtons', [
                 // if the user.email value is null, it means that he's not logged in
                 $rootScope.user.email = response.data.userEmail || null;
                 $rootScope.user.loggedin = $rootScope.user.email !== null;
+                $rootScope.user.is_staff= true;
```
edmorley: What's this for?  Was the "only show the cancel button on non-try if is_staff" feature broken before?

webapp/app/js/models/resultsets_store.js
```diff
@@ -852,16 +852,22 @@ treeherder.factory('ThResultSetStore', [
...
+                // if a job is completed, count the result instead of the state
+                if(job.state === "completed"){
                     memo[job.result]++;
                 }
+
+                // count these as coalesced, not as pending
+                if (job.job_coalesced_to_guid) {
+                    memo.coalesced++;
+                } else {
+                    memo[job.state]++;
```
edmorley: Do we still need to count the individual results? We're only using "still outstanding" vs "total", right? (with coalesced ignored from both groups)

webapp/app/js/models/resultsets_store.js
```diff
@@ -852,16 +852,22 @@ treeherder.factory('ThResultSetStore', [
...
-                    memo[job.state]++;
-                }else{
+                // if a job is completed, count the result instead of the state
+                if(job.state === "completed"){
                     memo[job.result]++;
                 }
+
+                // count these as coalesced, not as pending
+                if (job.job_coalesced_to_guid) {
+                    memo.coalesced++;
```
edmorley: Do we need to count coalesced?

webapp/app/partials/main/thResultCounts.html
```diff
@@ -1,10 +1,5 @@
...
-            <i class="fa fa-list fa-lg result-set-fa-icon"></i>
-            <span class="result-status-total-value"></span>
-    </span>
-    <span class="result-status-count-group">
-        <th-result-status-count ng-repeat="resultStatus in statusList"/>
-    </span>
+<span class="result-set-progress">
+    <span ng-if="percentComplete == 100">- Complete -</span>
+    <span ng-if="percentComplete < 100"
+          title="% of jobs completed vs. pending and running">{{percentComplete}}% - {{inProgress}} in progress</span>
```
edmorley: The percentage shown is "completed / total". The wording in the title to me reads as "pending+running / completed" or something along those lines.

For the title, how about this: "Proportion of jobs that are complete" :-)

edmorley:
Loving the look of the UI after this PR - awesome cleanup! :-)

webapp/app/css/treeherder.css
tojonmz: I just tried the branch (super nice!) but ya, I agree here. Perhaps it could float at the mid-point between the revision and the right hand UI elements, or something.

tojonmz:
Agreed, this is super beautiful @camd ! Particularly the new buttons minus their borders.

webapp/app/js/directives/persona.js
camd: dang, that was a mistake.  I did that just to see how it looked when in sheriff mode without having to re log in all the time.

edmorley: Ah makes sense! :-)

webapp/app/partials/main/thResultCounts.html
camd: I like it!  making it so...

webapp/app/css/treeherder.css
camd: Great point.  I tried a couple things and kind of like them to the right next to the buttons.  I'll go with that for now.

PR #355, opened 2015-02-04
Bug 1097090 - fix regression on jobs update

maurodoglio:


maurodoglio:
Please wait to review it, I found a couple of bugs I'd like to fix with this patch

edmorley:
I see there was an updated commit since your last message (Github doesn't send email notification for updated commits, only human comments), but the tests are failing, so not sure if I should be waiting longer or if you didn't see them? :-)

PR #358, opened 2015-02-06
Bug 1118253 - Deselect job when clicking any open area on main page

tojonmz:
This work fixes Bugzilla bug [1118253](https://bugzilla.mozilla.org/show_bug.cgi?id=1118253).

In it we support clearing a job, with a click anywhere in the resultset content area that isn't

* a link
* a resultset filter
* a resultset menu dropdown
* resultset button
* a job
* get-next 10 | 20 | 50 footer buttons

Otherwise, everywhere else on the main content area is fair game to clear the selected job, as one would expect it should be.

We basically place a clear on `th-content`, and then test the clicked element to see if it contains a `ignore-job-clear-on-click` attribute. If it doesn't have one, we clear the job.

Other than jobs, these areas highlighted, are the only areas *excluded* from clearing:

excluded

So we also fix Julien's workflow of clicking in the empty space below short pushes, similar to above.

Everything seems to be working correctly based on my local windows testing. I've marked some job clear code which I'd like to remove, and I believe is no longer needed. It used to handle clearing in the revision/job block region(only). It's flagged as such with `//` comments for visibility.

Once Ed has had a chance to check the behavior in general, I'd like to pull that code, squash so I just have one commit, and then after that potentially merge.

I've also pulled my previous stopgap 'get-next' footer fix, since a footer click now makes it to th-content. So it now clears without the previous fix.

I've read in some places that custom attrs should use the `data=' '` convention, rather than custom attr names, but I defer to others on that.

`closeJob()` is preserved with its usual behavior for *esc* and the job panel close *[x]* icon.

I may need to also rebase again, over camd's upcoming filter work in the resultset bar.

Tested on Windows:
FF Release **35.0.1**
Chrome Latest Release **40.0.2214.94 m**

Adding @camd for review, and @edmorley for testing.

### DIFF
--- a/webapp/app/js/directives/clonejobs.js
+++ b/webapp/app/js/directives/clonejobs.js
@@ -174,15 +174,6 @@ treeherder.directive('thCloneJobs', [
         }, 200);
     };
 
-    var clearJobCb = function(ev, el, job) {
-        clearSelectJobStyles();
-
-        // Reset selected job to null to initialize nav position
-        ThResultSetStore.setSelectedJob($rootScope.repoName);
-
-        $rootScope.$emit(thEvents.jobClear, job);
-    };
-
     var togglePinJobCb = function(ev, el, job){
         $rootScope.$emit(thEvents.jobPin, job);
     };
@@ -309,11 +300,6 @@ treeherder.directive('thCloneJobs', [
 
             ThResultSetStore.setSelectedJob($rootScope.repoName, el, job);
 
-        } else {
-            // If user didn't select a job or anchor clear the selected job
-            if (el.prop("tagName") !== "A") {
-                _.bind(clearJobCb, this, ev, el)();
-            }
         }
     };
 
--- a/webapp/app/js/providers.js
+++ b/webapp/app/js/providers.js
@@ -185,9 +185,6 @@ treeherder.provider('thEvents', function() {
             // fired (surprisingly) when a job is clicked
             jobClick: "job-click-EVT",
 
-            // fired on click outside a job element
-            jobClear: "job-clear-EVT",
-
             // fired when the job details are loaded
             jobDetailLoaded: "job-detail-loaded-EVT",
 
--- a/webapp/app/plugins/controller.js
+++ b/webapp/app/plugins/controller.js
@@ -284,11 +284,6 @@ treeherder.controller('PluginCtrl', [
             $rootScope.selectedJob = job;
         });
 
-        $rootScope.$on(thEvents.jobClear, function(event, job) {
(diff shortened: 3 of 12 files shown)


webapp/app/js/directives/clonejobs.js
```diff
@@ -174,14 +174,17 @@ treeherder.directive('thCloneJobs', [
...
 
-    var clearJobCb = function(ev, el, job) {
-        clearSelectJobStyles();
-
-        // Reset selected job to null to initialize nav position
-        ThResultSetStore.setSelectedJob($rootScope.repoName);
-
-        $rootScope.$emit(thEvents.jobClear, job);
-    };
+    /* *** Would like to remove this ***
```
camd: this seems safe to remove.

webapp/app/js/directives/clonejobs.js
```diff
@@ -309,12 +312,14 @@ treeherder.directive('thCloneJobs', [
 
             ThResultSetStore.setSelectedJob($rootScope.repoName, el, job);
 
-        } else {
-            // If user didn't select a job or anchor clear the selected job
-            if (el.prop("tagName") !== "A") {
-                _.bind(clearJobCb, this, ev, el)();
-            }
-        }
+        } /* else {
```
camd: this seems safe to remove, too.

webapp/app/js/providers.js
```diff
@@ -185,8 +185,10 @@ treeherder.provider('thEvents', function() {
             // fired (surprisingly) when a job is clicked
             jobClick: "job-click-EVT",
 
-            // fired on click outside a job element
-            jobClear: "job-clear-EVT",
+            /* *** Would like to remove this ***
```
camd: go for it

webapp/app/plugins/controller.js
```diff
@@ -280,10 +280,12 @@ treeherder.controller('PluginCtrl', [
             $rootScope.selectedJob = job;
         });
 
-        $rootScope.$on(thEvents.jobClear, function(event, job) {
-            $rootScope.selectedJob = null;
-            $scope.$digest();
-        });
+        /* *** Would like to remove this ***
```
camd: sgtm

camd:
This all looks good to me.  Go ahead and remove that code you asked about and I'll merge it.

tojonmz:
Cool, thanks @camd!

tojonmz:
@camd I've retested on both Firefox and Chrome rebased to latest master and with those proposed blocks removed, and the change still seems fine. I think we're ready to merge and test on dev.

PR #362, opened 2015-02-09
Bug 1124266 - Allow a larger max size for job details panel

tojonmz:
This work fixes Bugzilla bug [1124266](https://bugzilla.mozilla.org/show_bug.cgi?id=1124266).

In it, we allow the user a resizable job details panel, up to 75 percent of the content height. For the last half year or so, it had been fixed at 35 percent and not adjustable. While there, I also made the two dynamic content containers in the pinboard (pinboard content, related bugs) adjust on the fly to the user's panel size.

Here's the pinboard specifically, before and after. Some minor tweaks, generally the same but slight optimization in height for more content space:

pinboardcurrent

pinboardproposed

Here is the proposed default launch size of the job panel (unchanged):

jobpanelpinboarddefaultlaunchsize

Here is the maximum size of 75 percent, which requires manual interaction:

jobpanelpinboard75pct

And here is an example of a larger screen and content, in use at that size:

jobpanelpinboard75pctlarge

There may be some groovey flexbox way of ensuring the pinboard containers when empty fill their parent containers, but this approach seemed to work fine. I also did a bit of rearranging of the property orders.

Everything seems to be working on on windows, but if you could have a bash on other OS's that would be great. Dynamic resizing seems to work well for me.

Tested on Windows:
FF Release **35.0.01**
Chrome Latest Release **40.0.2214.111 m**

Adding @wlach for review and @camd and @lizzard for visibility.

### DIFF
--- a/webapp/app/css/treeherder.css
+++ b/webapp/app/css/treeherder.css
@@ -551,7 +551,7 @@ div#bottom-panel {
     font-size: 12px;
 
     height: 35%;
-    max-height: 35%;
+    max-height: 75%;
     flex: none;
     -webkit-flex: none;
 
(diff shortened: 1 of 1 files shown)


webapp/app/css/treeherder.css
```diff
@@ -836,36 +836,37 @@ ul.failure-summary-list li .btn-xs {
 }
 
 #pinned-job-list{
+    position: relative;
```
wlach: I'm not sure about your use of positioning here. What's the purpose of making this relative, and the changes below to absolute? Things seem to lay out fine without these.

wlach:
Most looks good (and works well) but I'm a bit concerned about the use of position: in the CSS. What is it necessary for?

tojonmz:
Yes, the use of `absolute` and `relative` ensures the child containers fill their parents even when there is no or little content, at any panel size (35% to 75%). Here's what you would get without them:

norelabs

That being said maybe there is a better implementation, but it's one I was able to get working well.

tojonmz:
Noting here for what it's worth
https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Flexible_boxes

*"Absolutely positioned children of a flex container are positioned so that their static position is determined in reference to the main start content-box corner of their flex container."*

I believe that applies to the `content` containers I've modified (pinned content, related-bug content) for each of their respective `relative` parent containers, which live in the pinboard itself.

Maybe I can use `flex-basis` though, or similar. I'll have a look.

tojonmz:
Oh, here's an SO also

wlach:
That makes a ton of sense! Ok, let's merge this in then.

PR #363, opened 2015-02-12
Bug 1131183 - Esc should close the graph tooltips

mishravikas:


webapp/app/js/perf.js
```diff
@@ -149,6 +149,11 @@ perf.controller('PerfCtrl', [ '$state', '$stateParams', '$scope', '$rootScope',
       $scope.$digest();
     };
 
+    Mousetrap.bind('escape', function() {
+      $scope.ttLocked = false;
+      $scope.hideTooltip();
+      });
```
tojonmz: I think maybe this closing brace can be indented 2 chars to the left? So it lines up under the M.

tojonmz:
Just noting, all the core platform shortcuts currently live in controllers/main, and we have an existing shortcut for `esc` (which closes all open windows, and deselects the active job). So we just want to make sure this new esc behavior in perf, happens only in the graph page context.

https://github.com/mozilla/treeherder-ui/blob/master/webapp/app/js/controllers/main.js

I haven't tried the branch, but I will do so if I can :)

@wlach I wonder if we want to add this function to the Help for discoverability? Or Perhaps a new Shortcuts block under the current one in Help, dedicated specifically to Perf shortcuts.

webapp/app/js/perf.js
wlach: Yes, please fix this up. :)

wlach:
@tojonmz Yeah, the perf stuff is completely seperate code, at least for now. We don't use any of the same controller files, so this shouldn't effect treeherder at all.

PR #365, opened 2015-02-13
Status count code cleanup

camd:
This updates the fontawesome icon library and removes some dead code that SHOULD have been removed when I removed the status counts.


edmorley:
This looks great! The changelog for FontAwesome 4.0.3 -> 4.3.0 is pretty large - there have been quite a few new additions - we may find some of the new icons come in handy later :-)

PR #366, opened 2015-02-14
Bug 1113281 - Update retrigger logic to always hit new retrigger endpoint

lightsofapollo:


webapp/app/js/models/job.js
```diff
@@ -84,6 +84,18 @@ treeherder.factory('ThJobModel', [
         });
     };
 
+    ThJobModel.retrigger = function(repoName, pk, config) {
+        // a static method to retrieve a single instance of ThJobModel
```
camd: looks like this comment is incorrect

webapp/app/plugins/controller.js
```diff
@@ -235,25 +231,43 @@ treeherder.controller('PluginCtrl', [
...
-                return null;
             }
         };
 
         $scope.retriggerJob = function() {
-            thBuildApi.retriggerJob($scope.repoName, getRequestId());
+            ThJobModel.retrigger($scope.repoName, $scope.job.id).then(function() {
+                var requestId = getBuildbotRequestId();
+                if (requestId) {
+                    return thBuildApi.retriggerJob($scope.repoName, requestId);
```
camd: I'm not quite clear on why you call ``ThJobModel.retribber`` and then also call ``thBuildApi.retriggerJob``.  Wouldn't that do a retrigger for on buildbot jobs twice?

lightsofapollo:
@camd After looking at this again I see I missed part of the cancel case fixing...

webapp/app/plugins/controller.js
```diff
@@ -235,25 +231,49 @@ treeherder.controller('PluginCtrl', [
...
+              thNotify.send(message, "danger", true)
+            });
         };
 
         $scope.cancelJob = function() {
-            thBuildApi.cancelJob($scope.repoName, getRequestId()).then(function() {
-                ThJobModel.cancel($scope.repoName, $scope.job.id);
-            });
+            var requestId = getBuildbotRequestId();
+            if (requestId) {
```
camd: This looks like it's the inverse of the retrigger.  Should we call the ``ThJobModel.cancel`` first and then the ``thBuildApi.cancelJob`` in the ``then`` section if we find the ``requestId``?

webapp/app/plugins/controller.js
```diff
@@ -235,25 +231,49 @@ treeherder.controller('PluginCtrl', [
...
+                  ThJobModel.cancel($scope.repoName, $scope.job.id);
+              });
+            } else {
+              thNotify.send(
+                "Unable to get request id for retrigger/cancel", "danger", true
+              );
+            }
         };
 
         $scope.cancelAll = function(resultsetId) {
```
camd: is there any concept of the pulse cancel for a group of jobs like this for the result set?

lightsofapollo: No one job = one cancel (cancel all handles this as sending a one message per job that was cancelled on the server side)

webapp/app/plugins/controller.js
lightsofapollo: updated now sorry for confusion here

camd:
@lightsofapollo - This looks good to me.  I take it you tested it and it works ok?  :)  
Are you able to merge this yourself?  Or should I merge it?

lightsofapollo:
@camd Going to give it another check and then merge

edmorley:
I'm wondering if we should hold off merging any more changes greater than a few lines until we get this deploy working. We're already starting to get pile-on effect with later additions causing more regressions.

PR #368, opened 2015-02-16
Bug 1121631 - Implement moustrap shortcuts

tojonmz:
This work fixes Bugzilla bugs [1121631](https://bugzilla.mozilla.org/show_bug.cgi?id=1121631) and [1118342](https://bugzilla.mozilla.org/show_bug.cgi?id=1118342).

This adds a few things beyond previously reverted PR https://github.com/mozilla/treeherder-ui/pull/338. Specifically it:

* addresses the slowdown in toggle unclassified
* uses `$scope.$evalAsync` for better optimization where formerly `$digest` was used
* ignores shortcuts on manual entry of classification comments with the `allowKeys()` function

An article on.Async can be found [here](http://www.bennadel.com/blog/2605-scope-evalasync-vs-timeout-in-angularjs.htm).

I've also left in some placeholders for next/previous job navigation, for @dewgeek's upcoming PR. In the event mine lands first, they can just be un-commented (and tested).

Some shortcuts as before, did not require digest or Async. Only those which needed them received them.

Everything seems to be working fine for all shortcuts in Firefox and Chrome, which the exception of toggle In-Progress, which is known to be not working at this time. I added an Async for it regardless for when it is fixed.

We'll want sheriffs to give this thorough testing on dev first, before pushes to stage or prod.

Tested on Windows:
FF Release **35.0.1**
Chrome Latest Release **40.0.2214.111 m**

### DIFF
--- a/webapp/app/js/controllers/main.js
+++ b/webapp/app/js/controllers/main.js
@@ -48,10 +48,17 @@ treeherder.controller('MainCtrl', [
             }
         };
 
-        $scope.processKeyboardInput = function(ev){
+        // Disable single key shortcuts in specified shortcut events
+        $scope.allowKeys = function() {
+            Mousetrap.unbind(['i', 'j', 'n', 'k', 'p', 'space', 'u', 'r', 'c']);
+        };
 
-            // If the user is in an editable element or the user is pressing
-            // shift, then disable keyboard events
+        // Process shortcut events
+        $scope.processKeyboardInput = function(ev) {
+
+            /* If the user is in an editable element or pressing shift
+             * then disable keyboard events, unless otherwise enabled
+             * in inputs by the 'mousetrap' class in markup */
             var activeElement = document.activeElement;
             if (activeElement.tagName === 'INPUT' ||
                 activeElement.tagName === 'SELECT' ||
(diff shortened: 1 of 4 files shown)


webapp/app/js/controllers/main.js
```diff
@@ -48,10 +48,17 @@ treeherder.controller('MainCtrl', [
             }
         };
 
-        $scope.processKeyboardInput = function(ev){
+        // Disable single key shortcuts in specified shortcut events
+        $scope.allowKeys = function() {
+            Mousetrap.unbind(['i', 'j', 'n', 'k', 'p', 'space', 'u', 'r', 'c']);
```
tojonmz: Wondering it might be useful to annotate the unbind, with an adjacent comment. Since we will exceed that 80 char line length when we add `left` and `right` arrow shortcuts shortly. eg.
```
Mousetrap.unbind([
  'i',    // Toggle display in-progress jobs (pending/running)
  'j',    // Select next unclassified failure
  'n',    // Select next unclassified failure
```

If not now, perhaps an idea for later.

camd:
looks great!

PR #371, opened 2015-02-18
Improve the initial setup experience for new contributors to the UI

edmorley:
* The sample config is optional, so don't use it for tests …
We want the sample config to now point at the production service API to make the first-run experience for new contributors easier. However the tests use the sample config as part of the test run, so this breaks them. However since the sample config is now optional, we can just not use it at all during the tests to fix the failures.

* Improve the API back-end prefs in the sample local config …
It's now easier to switch between the API back-ends and harder to make typos. The default backend has also been changed to production (instead of using the same host), since local.conf.js is now optional, so if someone just wants to use the same host as the UI for the back-end, then they can just not create local.conf.js in the first place. But for those who _do_ want to use production (eg those using web-server.js), initial setup is now a single |cp sample.local.conf.js local.conf.js| rather than also having to edit the file afterwards.

* Clean up the treeherder-ui installation instructions

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,42 +1,54 @@
 Installation
 ============
 
-You can run the webserver locally.  For now, static data is loaded for testing
-and development.
+You can work on the UI without needing a VM, by using web-server.js.
+There are a few limitations, since URL rewriting is not supported (and so some links will be broken), but it works well enough for quick testing. For instructions on how to serve the UI with working URL rewriting, use the treeherder-service Vagrant instructions.
 
+Cloning the Repo
+----------------
+
+* Clone the `treeherder-ui repo`_ from Github.
 
 Requirements
 ------------
 
-* node.js: http://nodejs.org/download/
+* Node.js_
 
-Execution::
+Running the web-server
+----------------------
 
-    cd webapp
-    ./scripts/web-server.js
+* Open a shell, cd into the root of the repository you just cloned and type:
 
+  .. code-block:: bash
 
-Endpoints
----------
+     cd webapp
+     cp app/js/config/sample.local.conf.js app/js/config/local.conf.js
+     ./scripts/web-server.js
 
-Once the server is running, you can nav to:
+Viewing the UI
+--------------
 
-* Jobs list: http://localhost:8000/app/index.html?tree=Try#/jobs
-* Log Viewer: http://localhost:8000/app/logviewer.html
+Once the server is running, you can navigate to:
+`<http://localhost:8000/app/index.html>`_
 
+Configuration
+=============
+
+The sample configuration makes the UI load job/push data from the production service API. If you wish to test the UI against stage/dev's service instead, adjust ``thServiceDomain`` in the config file created as part of installation:
+``webapp/app/js/config/local.conf.js``
+
+If you wish to run the full treeherder-service Vagrant project (service + UI), remember to remove local.conf.js or else change ``thServiceDomain`` within it to refer to ``vagrant``, so the UI will use the local Vagrant service API.
 
 Running the unit tests
 ======================
 
 The unit tests run with Karma: http://karma-runner.github.io/0.8/config/configuration-file.html
 
-
 Requirements
 ------------
 
-* [node.js](http://nodejs.org/download/)
+* Node.js_
 * karma: ``sudo npm install -g karma``
-
 
 Execution::
 
@@ -44,7 +56,7 @@ Execution::
     ./scripts/test.sh
 
 Build
------
+=====
 * Install grunt ``sudo npm install grunt``
 * Install the ``devDependencies`` in ``package.json``
 * Run the following command in ``treeherder-ui``:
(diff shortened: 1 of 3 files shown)


docs/installation.rst
```diff
@@ -1,50 +1,56 @@
...
+     cp app/js/config/sample.local.conf.js app/js/config/local.conf.js
+     ./scripts/web-server.js
 
-Once the server is running, you can nav to:
+Viewing the UI
+--------------
 
-* Jobs list: http://localhost:8000/app/index.html?tree=Try#/jobs
-* Log Viewer: http://localhost:8000/app/logviewer.html
+Once the server is running, you can nav to:
```
tojonmz: Wonder if we maybe say 'navigate' here.

docs/installation.rst
```diff
@@ -1,50 +1,56 @@
...
 
+* Jobs list: `<http://localhost:8000/app/index.html#/jobs?repo=try>`_
+* Log Viewer: `<http://localhost:8000/app/logviewer.html>`_
```
tojonmz: When navigating to this guy without a revision, the Log Viewer looks a bit odd and might trigger new-user questions. Perhaps the first jobs link is enough?

logviewerlink

webapp/app/js/config/sample.local.conf.js
```diff
@@ -3,17 +3,19 @@
  * file, you can obtain one at http://mozilla.org/MPL/2.0/. */
 
 'use strict';
+var production = "https://treeherder.mozilla.org";
+var stage = "https://treeherder.allizom.org";
+var dev = "http://treeherder-dev.allizom.org";
+var same_host = "";
```
tojonmz: I was just curious about the rationale for `same_host` vs. `local` or `localhost` here, though I'm sure there is a compelling reason.

I had tried local and localhost as vars, pointing to stage (just as a test) and it worked fine also.

edmorley:
Agree on both counts, have pushed an update. I've just put the main homepage link instead of the Try link - we also need to fix the URLs when switching repos, but that;s another matter.

webapp/app/js/config/sample.local.conf.js
edmorley: The name is arbitrarily picked, and I originally used 'local', however it occurred to me that when Vagrant is used, it's not actually connecting to localhost, it's connecting to 192.168.33.10 via the hosts local.treeherder.mozilla.org entry. I didn't want to call it 'vagrant' since the empty string is also used for stage/prod. So same_host seemed more appropriate. Happy for suggestions though :-)

tojonmz: Hm, if it's always going to be vagrant, I wonder about `localvagrant`. Maybe that communicates both things. It just seemed the use of 'same' implied it was the same as something else (that is not known).

edmorley: It is always the same as something else - the same as the UI host name :-)
(See block comment below)
Though perhaps this needs to be re-ordered to be clearer.

Or we could just call it 'vagrant' and ignore prod/stage for now, since they just don't have local.conf.js at all (even though they use the empty string effectively too - which is why I'd made the variable more generic and called it same_host)

docs/installation.rst
```diff
@@ -57,22 +61,10 @@ This will create a ``dist`` directory in ``treeherder-ui`` where concatenated an
 Configuration
 =============
```
tojonmz: I'm wondering if this Configuration section might be helpful following the Running the Web Server part above, or if not, perhaps stay here but still provide some context on the sample.local.conf.js which was the origin of the file. Without that, looking at this section it refers to a file, local.conf.js that doesn't exist by default.

webapp/app/js/config/sample.local.conf.js
tojonmz: I might vote for 'vagrant' in that case, if we think it might result in less questions in channel?

docs/installation.rst
```diff
@@ -57,22 +61,10 @@ This will create a ``dist`` directory in ``treeherder-ui`` where concatenated an
...
-
-Copy the sample file to ``local.conf.js`` and make your custom settings.
-If you want to run the UI using the node web server shipped with it, you need to change this line::
-    
-    window.thServiceDomain = "";
-
-to::
-
-    window.thServiceDomain = "https://treeherder.mozilla.org";
+If you wish to run the full treeherder-service Vagrant project (service + UI), remember to remove local.conf.js or else change thServiceDomain within it to use |same_host|, so the UI will use the local Vagrant service API.
```
tojonmz: This |same_host| would need adjusting also if we switch the config, but more interestingly, when I view this file this phrase is linked, but just gives me an identical installation page.

edmorley: Done; and yeah it seems that clashes with the markup.

docs/installation.rst
edmorley: I've moved higher up & specifically referenced the "file created above", but I think it still should be separate from "running the server". The first section is a basic "how to get running with the defaults as fast as possible", the later configuration section is optional.

webapp/app/js/config/sample.local.conf.js
edmorley: Cool, changed :-)

tojonmz:
This all looks good to me, awesome work and improvements here @edmorley particularly the var additions in the config file. The basic workflow and config file works fine at least on windows, as the instructions describe.

PR #372, opened 2015-02-18
Bug 1132506 - Option to 'mute' series temporarily

mishravikas:


webapp/app/js/perf.js
```diff
@@ -238,6 +238,17 @@ perf.controller('PerfCtrl', [ '$state', '$stateParams', '$scope', '$rootScope',
       $scope.reload();
     };
 
+    $scope.hideSeries = function(signature){
+      var plotdata = $scope.plot.getData();
+      for (var i = 0; i < plotdata.length; i++) {
+              if (plotdata[i].thSeries.signature == signature) {
```
wlach: Please use 2-level indentation here.

webapp/app/js/perf.js
```diff
@@ -238,6 +238,17 @@ perf.controller('PerfCtrl', [ '$state', '$stateParams', '$scope', '$rootScope',
       $scope.reload();
     };
 
+    $scope.hideSeries = function(signature){
+      var plotdata = $scope.plot.getData();
+      for (var i = 0; i < plotdata.length; i++) {
```
wlach: You could replace this iteration logic with an underscore function like:

var datumToHide = _.find(plotdata, function(datum) { return datum.thSeries.signature == signature });
datumToHide.points.show = !datumToHide.points.show;

webapp/app/partials/perf/perfctrl.html
```diff
@@ -16,6 +16,7 @@
             <span class="sr-only">Remove series</span>
           </button>
           <table>
+            <input title="Show/Hide series" type="checkbox" id="showHideCheck" ng-click="hideSeries(series.signature)" checked>
```
wlach: I think we should try to align this element in the bottom right corner (floating), as is done in graphs.mozilla.org. As it is, it really distracts from the series data information which is what we want to call attention to. Could you adjust the CSS to make this happen? If you need help, feel free to ask either myself or @tojonmz

tojonmz: Trivial, but I think we've been trying to capitalize just the first letter for Title consistency, unless it's using a specific platform term. The other Show/hide I could think of is here as an example. I'm easy though, maybe we update that one instead.
https://github.com/mozilla/treeherder-ui/blob/2f7f9d92f5bcbdaf62d232447f787adbbbfa169e/webapp/app/partials/main/thWatchedRepoNavPanel.html#L18

webapp/app/css/perf.css
```diff
@@ -143,4 +143,8 @@ to { opacity: 0; }
 #scrollable-dropdown-menu .tt-dropdown-menu {
   max-height: 150px;
   overflow-y: auto;
-}
\ No newline at end of file
+}
+#showHideCheck{
```
wlach: Also this should be a class, not an id. Also, put it below .graph-legend and make the property only apply in its context:

#show-hide-check -> .graph-legend .show-hide-check.

wlach: Oh yeah, and for consistency use foo-bar-baz notation for css

webapp/app/partials/perf/perfctrl.html
```diff
@@ -25,6 +25,7 @@
               </td>
             </tr>
           </table>
+          <input title="Show/Hide series" type="checkbox" id="showHideCheck" ng-click="hideSeries(series.signature)" checked>
```
wlach: You'll need to update this after making the changes I suggest above.

PR #378, opened 2015-02-20
Bug 1134746 - Fix similar jobs selection

johnp:
Reset similar_job_selected to null so that the ng-if in the main.html discards the current detail pane and after the request finishes the detail pane gets added again with the new data.
If there's need to add a loading animation I can add that as well.


johnp:
The comment above is not up-to-date anymore. The new commit message is:

This patch removes the one-time-binding that otherwise prevents the change of similar_job_selected from propagating to the display.
Also format the duration field a little bit better, to show 'unknown' if a job is not finished yet.

camd:
Thanks @johnp - Appreciate the fix!

PR #382, opened 2015-02-20
Bug 1094466 - Separate scroll-panels for similar jobs panel

johnp:
This patch gets rid of overflowing for the bottom header and propagates down the content heights in a way, that the plugins can have proper scroll panes.
If we wouldn't explicitly inherit the height, it'd get calculated by its children, and therefore the scroll pane would overflow.
The change to pluginpanel.html is needed because otherwise we would have multiple div's with an explicitly set height, although their contents are display:none.

### DIFF
--- a/webapp/app/css/treeherder.css
+++ b/webapp/app/css/treeherder.css
@@ -659,6 +659,7 @@ div#bottom-panel .navbar.navbar-dark .navbar-nav > li.active a:focus {
 
 #bottom-left-top, #bottom-center-top {
     min-height: 33px;
+    overflow-y: auto;
 }
 
 .pinned-job-count {
@@ -717,20 +718,38 @@ div#bottom-panel .navbar.navbar-dark .navbar-nav > li.active a:focus {
 
 #bottom-center-bottom {
     max-height: calc(100% - 33px);
+    -webkit-flex: 1;
+    flex: 1;
+    display: -webkit-flex;
+    display: flex;
 }
 
 #bottom-center-bottom > * {
-    height: 100%;
+    -webkit-flex: 1;
+    flex: 1;
+    display: -webkit-flex;
+    display: flex;
+}
+
+#bottom-center-bottom > * > * {
+    -webkit-flex: 1;
+    flex: 1;
+    display: -webkit-flex;
+    display: flex;
 }
 
 #bottom-center-bottom > * > * > * {
-    height: 100%;
+    -webkit-flex: 1;
+    flex: 1;
+    display: -webkit-flex;
+    display: flex;
 }
 
 #bottom-center-panel ul.failure-summary-list{
-    padding-left: 0px;
-    height: 100%;
+    padding-left: 0;
     overflow: auto;
+    width: 100%;
+    margin-bottom: 0;
 }
 
 ul.failure-summary-list li {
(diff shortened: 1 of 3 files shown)


camd:
@johnp - This looks great and is good to merge, but would you please rebase on the latest master first to be sure to remove any conflicts?  This is a great improvement!  

Thanks!!  :)
-Camd

johnp:
I've rebased the patch, but testing locally shows a couple of new issues I'll have to get my head around. Sometimes when selecting another job from the similar jobs list the according scroll pane gets invalidated somehow and gets re-rendered after 100-200 msecs... I'll check if that is related to the one-time-binding issue getting fixed, because that's the only thing I have in mind that could have an effect here. Also it doesn't work in Chrome, which is kind of weird... The "height: 100%" doesn't make the child div have the same height as the parent div (read: containing block) there, although it is explicitly set. Could be because it's calculated using "calc()" and Chromium has a bug there...

I'll investigate the issues and report back.

camd:
@johnp - thanks!

johnp:
I've reduced the chrome-issue quite a bit and came up with a jsfiddle that shows the issue. I've already minimized the html pretty much, but just copied the css... if someone knows a tool that extracts all used CSS rules out of a .css file using a given HTML I'd be very grateful for that. On the other hand, one can easily see the used rules in the inspector, so I left it like that for now.

https://jsfiddle.net/nct8xafm/3/

The issue somehow appears with this two divs:
´´´
    <div id="bottom-center-bottom">
        <div class="ng-scope">
´´´
´´´
#bottom-center-bottom { max-height: calc(100% - 33px); }
#bottom-center-bottom > * { height: 100%; }
´´´

#bottom-center-bottom correctly calculates a height in Chrome and Firefox (227px).
The child correctly inherits the exact same height in Firefox, but Chrome behaves like the parents height is not set at all, and, because of that, ignores the rule and instead calculates the height based on its children.

If someone could validate this or has an idea whats going on I would greatly appreciate it.

johnp:
I've minimized the CSS and simplified the HTML:

https://jsfiddle.net/nct8xafm/7/

In Firefox a scroll-bar should appear; In Chrome the text get's truncated.

johnp:
So, now it's 3 divs + 3 css rules to reproduce the problem (+ 2 css rules to make it visible without inspector):

https://jsfiddle.net/nct8xafm/8/

johnp:
So, here's the update that adds Chrome compatibility without removing Firefox compatibility or messing up other parts of the bottom panel (which was harder than it looks like ^^).
At the end of my investigation and try-and-error coding I also finally stumbled upon the Chrome issues (as linked in the commit message): 
https://code.google.com/p/chromium/issues/detail?id=341310

Fwiw I also used the -webkit prefixes as done in the rest of the file, although at least Chrome doesn't need them anymore since 1'nd 1/2 years ago and I did not test IE compatibility (do we need this?).

tojonmz:
As far as I know from last June-ish Treeherder mtg discussions with @jeads and @camd we originally only intended to support Firefox and Chrome browsers. I think that is still the case.

Unrelated, but when this change gets merged we'll want Sheriffs to test it on dev straight away, and sign off on it before subsequent pushes to stage/prod. As a process, that helps prevents unnecessary rollbacks.

johnp:
Great, thanks for the info.
There are two more notes about this patch:

In the Talos-Tab multiple <span>s now have a horizontal (column) layout instead of a vertical (row).
I spot-checked and could not find a Job where multiple of these were present (and populated; there were was one additional empty one always there, which I removed by moving the ng-if one element up), and fixing this would involve specifying a flow-direction: column on the div above which would make the scroll-panes in the Similar-Jobs-Tab disappear in Firefox (but not in Chrome).

On my dirty Firefox profile - and not in a new one - clicking on a job in the Similar-Jobs-Tab sometimes makes greyed-out horizontal scroll-bars appear for a brief moment on a couple page-elements. This seems to be caused by some of my modified settings or add-ons, so I would rather get this landed and hope no-one else sees this, but if not, compare the configuration and find out what's causing this issue.

johnp:
So, the first note above (about the Talos-Tab) seems to be our issue:
https://bugzilla.mozilla.org/show_bug.cgi?id=1000957

camd:
I actually don't see the horizontal lines on FF.  I think this is good.  :)

camd:
And, yeah, we're not worrying about IE support at the moment.

camd:
@johnp Thanks!!

PR #386, opened 2015-02-25
Add graphene build support.

bhearsum:
Add support for Graphene builds. I also bumped the "other" group to 99 because it seems like it should always be last. Let me know if I should use 42 though.


edmorley:
Looks good to me - thanks for bumping other to 99 :-)

PR #387, opened 2015-02-25
Bug 1136738 - Add better error messages when server fails due to auth er...

lightsofapollo:
...rors r=camd


camd:
looks good to me!

PR #388, opened 2015-02-25
 	Bug 1108832 - Support summary series in the frontend

wlach:


webapp/app/js/perf.js
```diff
@@ -28,11 +28,15 @@ perf.factory('seriesSummary', ['seriesColors', function(seriesColors) {
     if (signatureProps.job_group_symbol === "T-e10s") {
       extra = " e10s";
     }
-    var testName = signatureProps.suite + " " + signatureProps.test +
+    var testName = signatureProps.test;
+    if (testName === undefined)
+      testName = "summary";
```
jmaher: it seems to be common in this code to use {} even if there is only one line of code following it.  Can we add that for this if condition?

webapp/app/js/perf.js
```diff
@@ -28,11 +28,15 @@ perf.factory('seriesSummary', ['seriesColors', function(seriesColors) {
...
+
+    var name = signatureProps.suite + " " + testName +
       " " + optionCollectionMap[signatureProps.option_collection_hash] + extra;
-    var signatureName =  testName;
-    return { name: signatureName, signature: signature, platform: platform,
-             testName: testName, projectName: projectName, color: seriesColors[number] };
+    var signatureName = name;
+    return { name: name, signature: signature, platform: platform,
+             projectName: projectName, color: seriesColors[number] };
```
jmaher: here we define signatureName but don't use it.  Previously in the return statement we had:
"name: signatureName" as well as "testName: testName"

I understand that name is what testName used to be, so lets just remove signatureName.

PR #391, opened 2015-02-26
Bug 1137425 - Fixed annotation classification deletion

dewgeek:
while updating ThResultSetModel in all the controllers, It is not updated in this file, because of which, we are unable to delete classifications .

### DIFF
--- a/webapp/app/plugins/annotations/controller.js
+++ b/webapp/app/plugins/annotations/controller.js
@@ -6,10 +6,10 @@
 
 treeherder.controller('AnnotationsPluginCtrl', [
     '$scope', '$rootScope', 'ThLog', 'ThJobClassificationModel', 'thNotify',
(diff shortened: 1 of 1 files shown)


tojonmz:
@dewgeek can you provide a more referenced commit message? eg. 'Bug 1137425 - Fixed annotation classification deletion" etc. Thanks.

dewgeek:
@tojonmz , sorry about that. :)

tojonmz:
No, actually we want the commit message itself to have that `Bug 1137425 - Fixed classification deletion` message, not just the title of the PR.

All commits with rare exception should be prefixed with `Bug nnnnnn - ` so we know what bug the commit was fixing, when we look at the git log history.

camd:
@dewgeek Looks like we can close this PR, as it was subsumed by the PR for Bug 1113396

PR #392, opened 2015-02-26
Bug 1113396 - Make unclassified jobs visible in only-unclassified mode

dewgeek:
after deleting classification , it's failure classification id is changed to 1 , which makes it unclassified error

### DIFF
--- a/webapp/app/plugins/annotations/controller.js
+++ b/webapp/app/plugins/annotations/controller.js
@@ -6,10 +6,10 @@
 
 treeherder.controller('AnnotationsPluginCtrl', [
     '$scope', '$rootScope', 'ThLog', 'ThJobClassificationModel', 'thNotify',
-    'thEvents', 'ThResultSetModel', 'ThBugJobMapModel', 'thTabs',
+    'thEvents', 'ThResultSetStore', 'ThBugJobMapModel', 'thTabs',
     function AnnotationsPluginCtrl(
         $scope, $rootScope, ThLog, ThJobClassificationModel,
-        thNotify, thEvents, ThResultSetModel, ThBugJobMapModel, thTabs) {
+        thNotify, thEvents, ThResultSetStore, ThBugJobMapModel, thTabs) {
 
         var $log = new ThLog(this.constructor.name);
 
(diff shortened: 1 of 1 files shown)


tojonmz:
Same here with the commit message, as with https://github.com/mozilla/treeherder-ui/pull/391#issuecomment-76292586. Probably something like: `Bug 1113396 - Make unclassified jobs visible in only-unclassified mode`

webapp/app/plugins/annotations/controller.js
```diff
@@ -21,6 +21,14 @@ treeherder.controller('AnnotationsPluginCtrl', [
         }, true);
 
         $scope.deleteClassification = function(classification) {
+
+            var key = "key" + classification.job_id;
+            var jobMap = ThResultSetStore.getJobMap($rootScope.repoName);
+            var job = jobMap[key].job_obj;
+
+            job.failure_classification_id = 1 ;
```
tojonmz: There no space needed after the `1` and the `;`

webapp/app/plugins/annotations/controller.js
```diff
@@ -21,6 +21,14 @@ treeherder.controller('AnnotationsPluginCtrl', [
...
+            ThResultSetStore.updateUnclassifiedFailureMap($rootScope.repoName, job)
```
tojonmz: I think this needs a trailing `;`

dewgeek:
@tojonmz , I squashed my changes to avoid multiple commits during merge

tojonmz:
@dewgeek cool, just a note that I think you'll typically only squash right at the end, once the main reviewer (eg. @camd) has approved the entire PR and all the changes. Although Github does show the outdated diffs in the conversation history, they lose the series of commits in the commits tab because of the squash. I believe normally you only want to blow those away with a squash once, right at the end, before the merge.

It could vary with with the repo owners though. It's just my general understanding of best practice.

tojonmz:
Oh, and the commit message still needs to be updated, per my first comment in the conversation above.

dewgeek:
@tojonmz , commit message ammended :)

camd:
looks good!  Works well.  thanks  @dewgeek!!  :)

PR #393, opened 2015-02-27
Bug 1133545 - Create a shortcut to focus the Filter field

tojonmz:
This work fixes Bugzilla bug [1133545](https://bugzilla.mozilla.org/show_bug.cgi?id=1133545).

This adds the **f** key as a shortcut to focus the "Filter platforms & jobs" field. In it we:

* similarly issue a click in the shortcut rather than focus, which selects the string if it exists
* create a new `clickThis` directive to support that behavior
* no longer re-invoke the filter as soon as the cell is empty, instead only after the user hits Enter
* create a global behavior which suppresses (makes transparent) *all* placeholder text on focus

The suppression/transparency of placeholder text on click was approved by Ryan in channel. I checked and it didn't seem to have any adverse impacts in the pinboard, etc. It seems to add clarity, but we could easily remove it if needed.

Invoking an empty filter by typing Enter, was also approved by Ryan in channel. This means there is one extra keystroke to clear a filter (backspace, *Enter*), but this makes both initial input, and re-input, consistent.

There are no changes in UI appearance but here's an example of a Filter shortcut invocation with an existing filter present:

shortcutreinvocation

Everything seems to be working well in all my local testing, and it seems a huge improvement over having to manually go and click on the filter input every time. Focus also returns to the main panel as soon as you've hit Enter, so the usual j,k,n,p navigation and other shortcut workflows seem to work seamlessly while switching back and forth between filter states.

The only scenario a bit ambiguous, but has always been on production, is:

* enter or empty a filter but don't hit Enter
* *click away* somewhere else

Now you have an empty filter field, but a filter state which doesn’t match since it hasn’t been invoked. However this is our current behaviour on production. I spent some time trying to get both behaviors working, but could only get one (click away) or the other (enter). The input doesn't appear to blur on a click away, so we can't do an `ng-blur=search($event)` and then check for it and invoke the search in filters.js. Perhaps for later.

I've updated the Help and also added the .unbind so we don't issue a filter command when typing text in other inputs which support multi-key shortcuts during text entry (eg. comments, add-related-bugs).

Tested on OSX 10.9.5:
FF Release **35.0.1**

### DIFF
--- a/webapp/app/js/controllers/main.js
+++ b/webapp/app/js/controllers/main.js
@@ -154,9 +154,7 @@ treeherder.controller('MainCtrl', [
                 // Prevent shortcut key overflow during focus
                 ev.preventDefault();
 
-                $scope.$evalAsync(
-                    $rootScope.$broadcast('click-this', "platform-job-text-search-field")
-                );
+                angular.element('#platform-job-text-search-field').trigger('click');
             });
 
             // Shortcut: escape closes any open panels and clears selected job
--- a/webapp/app/js/controllers/main.js
+++ b/webapp/app/js/controllers/main.js
@@ -50,7 +50,7 @@ treeherder.controller('MainCtrl', [
 
         // Disable single key shortcuts in specified shortcut events
         $scope.allowKeys = function() {
-            Mousetrap.unbind(['i', 'j', 'n', 'k', 'p', 'space', 'u', 'r', 'c']);
+            Mousetrap.unbind(['i', 'j', 'n', 'k', 'p', 'space', 'u', 'r', 'c', 'f']);
         };
 
         // Process shortcut events
@@ -137,7 +137,6 @@ treeherder.controller('MainCtrl', [
                         $rootScope.$emit(thEvents.jobPin, $rootScope.selectedJob)
                     );
 
-
                     // Prevent shortcut key overflow during focus
                     ev.preventDefault();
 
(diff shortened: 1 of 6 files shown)


webapp/app/js/controllers/main.js
```diff
@@ -150,6 +149,16 @@ treeherder.controller('MainCtrl', [
                 }
             });
 
+            // Shortcut: enter a custom job or platform filter
+            Mousetrap.bind('f', function(ev) {
+                // Prevent shortcut key overflow during focus
+                ev.preventDefault();
+
+                $scope.$evalAsync(
+                    $rootScope.$broadcast('click-this', "platform-job-text-search-field")
```
camd: I found this on stack overflow.  I wonder if it would negate the need for the new directive:
http://stackoverflow.com/questions/22447374/how-to-trigger-ng-click-angularjs-programmatically
``angular.element('#myselector').trigger('click');``

I haven't tried this approach, so maybe there's a reason not to do that.  Just wanted to throw it out there.

camd: Yeah, I played around with this and looks like this works:

    angular.element('#platform-job-text-search-field').trigger('click')

And then I think you don't need the directive in ``directives/main.js``

camd:
This is a really cool feature.  Would you mind playing with the code line I gave you and see if it works for you, too?  Thanks!!!

tojonmz:
Dig, agreed. I'll try it out and make sure it's groovy here also and report back. Thanks @camd!

webapp/app/js/controllers/main.js
tojonmz: Totally agreed on this approach. I had searched for an S.O. on this and also dug through the angular docs but had missed the method. It's one of those that surely had to be there.

tojonmz:
@camd did the interaction performance change for you from the (now unused) $scope.$evalAsync? I don't think I detected delays here using the new `angular.element` approach, but my machine is so fast it might mask differences.

I tested all the functionality with the new approach, and in all functionality respects it seems to be working fine.

camd:
I don't think it needs the $evalAsync because it's put into the $digest cycle by the angular click event.  But, yeah, worked great for me locally.  :)

PR #395, opened 2015-03-02
Document the new spidermonkey builds

hotsphink:
https://bugzilla.mozilla.org/show_bug.cgi?id=1136309

### DIFF
--- a/webapp/app/help.html
+++ b/webapp/app/help.html
@@ -167,6 +167,12 @@
                        SpiderMonkey</a> ARM Simulator Build
                   </td>
                 </tr>
+                <tr><th>p</th>
+                  <td>
+                    <a href="https://wiki.mozilla.org/Javascript:Automation_Builds">
+                       SpiderMonkey</a> Plain Shell Build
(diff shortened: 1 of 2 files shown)


tojonmz:
If we could pre-pend the bugnumber to the commit message before landing that would be nice, if it's not too big a hassle? I think you could squash it into your current commit if you do that.
eg. `Bug 1136309 - Document the new spidermonkey builds`

It just makes it a bit easier for historical purposes.

Thanks for helping out Steve!

hotsphink:
Ok, I will try to figure out what that means and how to do that. (Sorry, git and github newbie.)

tojonmz:
Oh, you can leave the PR open no problems. You actually don't really need to squash in this case. You could just amend your existing commit and force the push to your branch. It will be reflected here.

eg. locally in your git client:
`git commit --amend -m "Bug 1136309 - Document the new spidermonkey builds"`
`git push origin F-new-SM-builds --force`

Assuming the above `origin` is your remote name.

PR #396, opened 2015-03-02
Bug 1136309 - Document the new spidermonkey builds

hotsphink:
Is this what I'm supposed to do? I force-pushed an amended commit, then made a new PR with it.

### DIFF
--- a/webapp/app/help.html
+++ b/webapp/app/help.html
@@ -167,6 +167,12 @@
                        SpiderMonkey</a> ARM Simulator Build
                   </td>
                 </tr>
+                <tr><th>p</th>
+                  <td>
+                    <a href="https://wiki.mozilla.org/Javascript:Automation_Builds">
+                       SpiderMonkey</a> Plain Shell Build
+                  </td>
+                </tr>
                 <tr><th>d</th>
                   <td>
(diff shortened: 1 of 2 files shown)


tojonmz:
Sure, this is ok for this instance. Normally though you can leave the first PR open (https://github.com/mozilla/treeherder-ui/pull/395) and just force push to your remote (origin). Your --amend would be reflected in that PR.

For now, this one time is ok though.

That way we don't end up with tons of closed PR's which didn't result in any code changes.

tojonmz:
Oh, yes and you can find it on IRC in #treeherder if you have any questions, too.

tojonmz:
Oh, and I see @edmorley r+'d the original attachment to the bug, but if you want a second check, you can:

* paste the url to this PR as a 'text attachment' to the bug
* title the attachment as something... eg. "treeherder-ui-PR#396"
* set Ed as the reviewer again on that attachment

I'll defer to Ed until we hear back just to be safe. I'm just not sure if there have been changes since his patch review.

edmorley:
I'm fine to merge now :-)

tojonmz:
Thanks Ed!

PR #399, opened 2015-03-04
Bug 1113873 - Support existence of struct log summary artifact

camd:
When the Service is generating the structured log summary artifact, the
UI needs to handle that it will come back in the
endpoint for log artifacts. This change will just ignore the new
artifact.

### DIFF
--- a/webapp/app/plugins/controller.js
+++ b/webapp/app/plugins/controller.js
@@ -125,7 +125,9 @@ treeherder.controller('PluginCtrl', [
                         }, []);
                     }
                     // the fourth result comes from the jobLogUrl artifact
-                    $scope.job_log_urls = results[3];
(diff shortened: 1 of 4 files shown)


webapp/app/plugins/controller.js
```diff
@@ -124,7 +124,9 @@ treeherder.controller('PluginCtrl', [
                         }, []);
                     }
                     // the fourth result comes from the jobLogUrl artifact
-                    $scope.job_log_urls = results[3];
+                    // filter to only the builds-4h logs
+                    $scope.job_log_urls = _.where(results[3], {name: 'builds-4h'});
```
edmorley: We'll need to change the name here if we make the s/build-4h/.../ change - but for now looks great :-)

PR #400, opened 2015-03-04
Bug 1139301 - Make the date-format consistent and centralized

tojonmz:
This fixes Bugzilla bug [1139301](https://bugzilla.mozilla.org/show_bug.cgi?id=1139301).

This switches the date format for the job details panel for Request, Start, and End, from **m/d/yy**, to the format highlighted below.

In addition to removing ambiguity for all users (folks who read **m/d** or **d/m**), it makes it consistent with the resultset header. We use 24hour time, but we could add an AM/PM here if we feel it is really needed.

I'm awaiting feedback from @darktrojan just to make sure he's happy with it, before we merge.

jobdetaildateformat

Tested on OSX 10.9.5:
FF Release **35.0.1**

Adding @edmorley for review and @darktrojan for visibility.

### DIFF
--- a/webapp/app/plugins/controller.js
+++ b/webapp/app/plugins/controller.js
@@ -6,13 +6,13 @@
 
 treeherder.controller('PluginCtrl', [
     '$scope', '$rootScope', '$location', 'thUrl', 'ThJobClassificationModel',
-    'thClassificationTypes', 'ThJobModel', 'thEvents', 'dateFilter',
+    'thClassificationTypes', 'ThJobModel', 'thEvents', 'dateFilter', 'thDateFormat',
     'numberFilter', 'ThBugJobMapModel', 'thResultStatus', 'thJobFilters',
     'ThResultSetModel', 'ThLog', '$q', 'thPinboard', 'ThJobArtifactModel',
     'thBuildApi', 'thNotify', 'ThJobLogUrlModel', 'thTabs', '$timeout',
     function PluginCtrl(
         $scope, $rootScope, $location, thUrl, ThJobClassificationModel,
-        thClassificationTypes, ThJobModel, thEvents, dateFilter,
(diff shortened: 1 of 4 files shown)


edmorley:
Thank you for doing this :-)
I'm wondering if we should move the format string somewhere shared?
With these additions, we're now using the same string in at least 5 places.
I don;t know if https://github.com/mozilla/treeherder-ui/blob/master/webapp/app/js/values.js is the best place, or somewhere else - @camd ?

tojonmz:
That sounds like a neat idea, eg. `thDateFormat` or similar.

tojonmz:
@edmorley @camd so I did that variant using values.js, per the above `- values.js` commit. We can get 4 of the 6 dates, the other two use the date filter directly in the markup, so I just set them so they are consistent.

Ryan has approved the date changes across the various UI elements, in channel.

dateformat

If you think it looks good let me know and I'll squash before merge.

webapp/app/plugins/controller.js
```diff
@@ -6,13 +6,13 @@
 
 treeherder.controller('PluginCtrl', [
     '$scope', '$rootScope', '$location', 'thUrl', 'ThJobClassificationModel',
-    'thClassificationTypes', 'ThJobModel', 'thEvents', 'dateFilter',
+    'thClassificationTypes', 'ThJobModel', 'thEvents', 'dateFilter', 'thDateFormat',
```
tojonmz: I chose to go 84 chars here, rather than open a new line.

tojonmz:
Ed has r+'d this in the bug, so I will squash and merge.

tojonmz:
Referencing also an earlier PR (https://github.com/mozilla/treeherder-ui/pull/263) in case we ever decide we want relative time for some elements. I had provided a potential format at the bottom of this [comment](https://github.com/mozilla/treeherder-ui/pull/263#issue-48048722).

PR #401, opened 2015-03-04
Bug 1139599 - Option to highlight revision in perfherder

mishravikas:
@wlach I don't think we have a bug filed for this on bugzilla.

### DIFF
--- a/webapp/app/js/perf.js
+++ b/webapp/app/js/perf.js
@@ -110,6 +110,13 @@ perf.controller('PerfCtrl', [ '$state', '$stateParams', '$scope', '$rootScope', 
       $scope.$digest();
 
       this.plot.unhighlight();
+
+      //Highlight the points persisted in the url
+      $scope.seriesList.forEach(function(series, i) {
+        if (series.highlighted.length > 0 && series.visibility) {
+            $scope.plot.highlight(i, series.highlighted[0]);   
+        }
+      });
       this.plot.highlight(s, item.datapoint);
     };
 
@@ -147,6 +154,13 @@ perf.controller('PerfCtrl', [ '$state', '$stateParams', '$scope', '$rootScope', 
         $scope.ttHideTimer = setTimeout(function() {
           $scope.ttHideTimer = null;
           $scope.plot.unhighlight();
+
+          //Highlight the points persisted in the url
+          $scope.seriesList.forEach(function(series, i) {
+            if (series.highlighted.length > 0 && series.visibility) {
+                $scope.plot.highlight(i, series.highlighted[0]);   
+            }
+          });
           tip.animate({ opacity: 0, top: '+=10' },
                       250, 'linear', function() {
                         $(this).css({ visibility: 'hidden' });
@@ -163,6 +177,7 @@ perf.controller('PerfCtrl', [ '$state', '$stateParams', '$scope', '$rootScope', 
     Mousetrap.bind('escape', function() {
       $scope.ttLocked = false;
       $scope.hideTooltip();
+      $scope.plot.unhighlight();
     });
 
     function plotGraph() {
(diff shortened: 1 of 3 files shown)


webapp/app/js/perf.js
```diff
@@ -265,6 +295,38 @@ perf.controller('PerfCtrl', [ '$state', '$stateParams', '$scope', '$rootScope',
       $scope.reload();
     }
 
+    $scope.findRevision = function(){
+      var rev = $scope.higlightRevision;
+      if (rev.length == 12) {
+        var plotdata = $scope.plot.getData();
+        plotdata.forEach(function(series, i) {
+          if (series.points.show) {
+            var url = thServiceDomain + "/api/project/" + series.thSeries.projectName + "/resultset/?format=json&revision=" + rev + "&with_jobs=false";
```
tojonmz: I wonder if these guys can be wrapped at the ~80 char mark.

webapp/app/js/perf.js
```diff
@@ -110,6 +110,13 @@ perf.controller('PerfCtrl', [ '$state', '$stateParams', '$scope', '$rootScope',
       $scope.$digest();
 
       this.plot.unhighlight();
+
+      //Highlight the points persisted in the url
+      $scope.seriesList.forEach(function(series, i) {
+        if (series.highlighted.length > 0 && series.visibility) {
```
wlach: Change this to:

if (series.highlighted && ...)

That will ensure old urls still work.

Also, please factor this out into a common function like highlightDatapoints and call it from everywhere you're using, instead of duplicating this block. :)

webapp/app/js/perf.js
```diff
@@ -110,6 +110,13 @@ perf.controller('PerfCtrl', [ '$state', '$stateParams', '$scope', '$rootScope',
       $scope.$digest();
 
       this.plot.unhighlight();
+
+      //Highlight the points persisted in the url
```
wlach: Leave a space between '//' and 'Highlight' i.e. //Highlight the points persisted in the url -> // Highlight the points persisted in the url

webapp/app/partials/perf/perfctrl.html
```diff
@@ -35,8 +35,11 @@
   <div id="graphs">
     <select ng-model="myTimerange" ng-options="timerange.text for timerange in timeranges track by timerange.value" ng-change="timeRangeChanged()">
     </select>
-    <hr/>
     <div id="graph"></div>
+    <div id="graph-bottom">
```
wlach: I think we should make displaying this conditional on seriesList have a length greater than 0. Show be able to do this with ngShow: https://docs.angularjs.org/api/ng/directive/ngShow

webapp/app/js/perf.js
```diff
@@ -222,6 +237,20 @@ perf.controller('PerfCtrl', [ '$state', '$stateParams', '$scope', '$rootScope',
...
+      $scope.seriesList.forEach(function(series, i) {
+        if (series.highlighted.length > 0 && series.visibility) {
+            showResetRevButton = true;
+            $scope.plot.highlight(i, series.highlighted[0]);
+            $('#findRevText').val(series.highlighted[1]);   
+        }
+      });
+
+      if (showResetRevButton) {
+        $(".resetHighlightButton").show();
```
wlach: It would be more angular-like to have a variable like $scope.showResetRevButton, then make displaying the button conditional on it being true using the ngShow directive:  https://docs.angularjs.org/api/ng/directive/ngShow

webapp/app/js/perf.js
wlach: Should be :)

mishravikas:
@wlach @tojonmz can you please close this pr? I've opened a new one because of very messy merge conflicts.

wlach:
Sure thing

PR #402, opened 2015-03-04
Bug 1139565 - Add more polish to the Help build/job tables

tojonmz:
This work fixes Bugzilla bug [1139565](https://bugzilla.mozilla.org/show_bug.cgi?id=1139565).

In it we improve and make a more direct association with what the users see in Help, with what they see in the platform. Here's the before:

jobcodesbefore

And proposed:

jobcodesproposed

I used a style evocative of the running state/color, but slightly less dark, so the columns don't compete so much with their adjacent text. But so they are still readable. I'd be interested to know how that luminance looks on other folks machines to make sure it's ok.

Tested on OSX 10.9.5:
FF Release **35.0.1**

### DIFF
--- a/webapp/app/css/treeherder.css
+++ b/webapp/app/css/treeherder.css
@@ -1753,6 +1753,25 @@ fieldset[disabled] .btn-repo.active {
     font-weight: normal;
 }
 
+.help-btn-sm tr > th > button {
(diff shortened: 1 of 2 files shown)


tojonmz:

PR #403, opened 2015-03-04
Bug 1089737 - Unbind left/right navigation in mousetrap inputs

tojonmz:
This is a supplemental fix to Bugzilla bug [1089737](https://bugzilla.mozilla.org/show_bug.cgi?id=1089737).

This unbinds the left/right keys when in any mousetrap enabled input, for example add-related-bugs or classification-comment. Without the unbind, if the user uses the left or right arrows to traverse text in those inputs, the selected job changes.

Here's an example screen grab of that workflow:

arrowkeyrelatedbug

I did some tidying while I was there, and a planned annotation of the unbound keys now that we have exceeded the styleguide max line length.

Everything seems to be working fine in both browsers.

Tested on OSX 10.9.5:
FF Release **35.0.1**

Adding @wlach for review.

### DIFF
--- a/webapp/app/js/controllers/main.js
+++ b/webapp/app/js/controllers/main.js
@@ -50,7 +50,20 @@ treeherder.controller('MainCtrl', [
 
         // Disable single key shortcuts in specified shortcut events
         $scope.allowKeys = function() {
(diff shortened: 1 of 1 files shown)


tojonmz:
I did some further integration/sanity testing this morning just to re-check prior, and we seem to be fine. @wlach has r+d this in the bug, so I will merge.

PR #407, opened 2015-03-06
 	Bug 1140434 - Don't compile local.conf.js into minified perf.js

wlach:


webapp/app/perf.html
```diff
@@ -28,7 +30,6 @@
   <script src="vendor/jquery.flot.js"></script>
   <script src="vendor/jquery.flot.time.js"></script>
   <script src="vendor/mousetrap.min.js"></script>
-  <script src="js/config/local.conf.js"></script>
```
edmorley: For files other than perf.html, we move this outside of the build block, instead of removing entirely, eg:
https://github.com/mozilla/treeherder-ui/blob/master/webapp/app/index.html#L108

This means someone using it locally can still modify the paths + when we do bug 1131240, all pages will get that for free (we'll use local.conf.js on stage/prod for the new relic snippet).

wlach: Uh, isn't that what I just did? :)

edmorley: Uh oops. Sunlight on screen + viewing angle slightly off meant I couldn't see the green. Ignore me :-)

PR #408, opened 2015-03-07
Bug 1140734 - Add more polish to logviewer log lines

tojonmz:
This work fixes Bugzilla bug [1140734](https://bugzilla.mozilla.org/show_bug.cgi?id=1140734).

In it we convert the log lines to fixed-width, and hopefully add some subtle readability improvements. Here's the before:

loglinescurrent

Here's the after:

fixedwidthlogproposed

And I went with a slightly stronger red to help the errors pop a bit better than they did before:

errorcolorproposed

I also reduced the log line text size to small. I am hopeful that is still readable on all systems, since it gives us more content per page.

I have some other enhancement ideas, but we can leave those for later.

I had to make a css addition to the load help message "Click a log step above to view", so it remains sans serif. During the color change for errors I removed a css reference to `lv-line-number` since we aren't currently drawing them and that class doesn't appear to be referenced in the repo.

I checked in with @KWierso in channel with the above and the change seemed to be well received. Perhaps we can put it up on dev/stage after review iterations, and let the sheriffs evaluate it there.

Tested on OSX 10.9.5:
FF Release **36.0.1**

Adding @camd for review and @edmorley for visibility.

### DIFF
--- a/webapp/app/css/logviewer.css
+++ b/webapp/app/css/logviewer.css
@@ -38,17 +38,23 @@ body {
     width: 100%;
     overflow: auto;
     bottom: 0;
-    padding: 20px 15px 0;
+    font-family: monospace;
+    font-size: small;
     white-space: nowrap;
-    background: #EFEFEF;
+    background: #f8f8f8;
+}
+
+.lv-log-container > div:nth-child(even) {
+    background: #fff;
 }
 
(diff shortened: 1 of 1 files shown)


edmorley:
Looks great - thank you! :-)

PR #409, opened 2015-03-08
Bug 1136232 - More gracefully handle disabled sessionStorage

edmorley:
Some users set dom.storage.enabled to false to disable local storage for privacy reasons. Let's handle this more gracefully. In the future when we use local storage for more features, we'll want to break this out to somewhere common and use a shim of some sorts, but for now this will do.


edmorley:
Have updated the PR slightly, see the individual commits for clearer diffs :-)

PR #410, opened 2015-03-08
Bug 1102851 - Relinkify logviewer revision and reduce run-data font

tojonmz:
This change is supplemental to broader work requested in Bugzilla bug [1102851](https://bugzilla.mozilla.org/show_bug.cgi?id=1102851).

Nothing major here, but for some reason, I think accidental, we unlinked the logviewer run-data revision SHA in this commit https://github.com/mozilla/treeherder-ui/commit/c7d10090a01e13f11438fd60ffe3b470e4f7a15d . I wanted to re-link it. While I was there I reduced the font sizes in the upper data blocks. It makes more real estate for the log lines below, and reduces the frequency of vertical overflow in the first data block.

Here's the before and after:

rundatacurrent

rundataproposed

It seems to still be readable in testing on a lower res Windows machine and on my retina mbp.

There are other issues to be sorted (among many) but this addresses some small annoyances and will dovetail with work in [PR408](https://github.com/mozilla/treeherder-ui/pull/408), which I'd like us to land first.

After that I can rebase this on top of PR408  before potential merge. I may be able to economize the css changes here, depending on the font size changes landed in 408.

Tested on OSX 10.9.5:
FF Release **36.0.1**

Adding @camd for review and @edmorley for visibility.

### DIFF
--- a/webapp/app/css/logviewer.css
+++ b/webapp/app/css/logviewer.css
@@ -66,7 +66,9 @@ body {
     background-color: #FFF;
     padding: 20px 5px 4px;
     display: table-row;
+    font-size: 12px;
 }
+
 .run-data .break-word{
     -ms-word-break: break-all;
     word-break: break-all;
@@ -95,6 +97,7 @@ body {
     -o-user-select: text;
     user-select: text;
     white-space: normal;
+    font-size: 12px;
 }
 
 .logviewer-step.selected {
@@ -128,12 +131,10 @@ body {
 
 .logviewer-stepbar .raw-log {
     padding-right: 5px;
-    font-size: 12px;
 }
 
 .logviewer-stepbar input[type="checkbox"] {
     margin: 0;
-    vertical-align: middle;
 }
 
 .lv-line-no.label {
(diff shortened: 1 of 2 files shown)


tojonmz:
So @camd @edmorley I've rebased and re-pushed, Ryan has signed off on the size reduction, so I think we're good to review now.

webapp/app/css/logviewer.css
```diff
@@ -95,6 +97,7 @@ body {
     -o-user-select: text;
     user-select: text;
     white-space: normal;
+    font-size: 12px;
```
tojonmz: We need to override here its .btn inheritance (14px), otherwise I would have set 12 only on .run-data.

webapp/app/css/logviewer.css
```diff
@@ -146,7 +147,7 @@ body {
     padding: 26px 0 0 20px;
     font-weight: bold;
     font-family: sans-serif;
-    font-size: 14px;
+    font-size: 12px;
```
tojonmz: Same here.

PR #411, opened 2015-03-08
Bug 1134840 - Use the current result set filters when using "get next"

edmorley:
The second argument to fetchResultSets() determines whether the current result set filters are preserved for the call to fetch more result sets. Previously the filters were not preserved, causing unwanted result sets to be returned (eg pushes by other authors, when the author filter was being used).


tojonmz:
This looks great to me testing this branch on both Firefox and Chrome with get-next 10|20|50 with single and multiple filters (eg. author= plus filterSearcStr=) on OS X 10.9.5.

PR #412, opened 2015-03-09
Bug 918281 - Add dedicated links to tree rules in the tree info pane

KWierso:
This adds a dedicated section for tree rules to the tree info pane for any given repository, defaulting to a generic tree rules page in the case where there's no specific page for the current repository.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/412)

### DIFF
--- a/webapp/app/js/directives/top_nav_bar.js
+++ b/webapp/app/js/directives/top_nav_bar.js
@@ -100,6 +100,11 @@ treeherder.directive('thWatchedRepoInfoDropDown', [
                     scope.message_of_the_day = newVal.message_of_the_day;
                 }
             }, true);
+            if (attrs.rules) {
+              scope.treeRules = attrs.rules;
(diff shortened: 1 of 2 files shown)


tojonmz:
I just tested this branch locally on OSX 10.9.5 using Firefox and Chrome and the link seems fine. I gather unsupported repos will still have this link (eg. Bugzilla) although they will point to the Firefox tree rules? Those repos do contain a "*repo* is not supported in treestatus.mozilla.org" message however, which should help.

edmorley:
Mass-closing older inactive PRs.

For this issue, either changes will need to be first made upstream in TreeStatus, or else the bug might end up being wontfixed, depending on the outcome of https://bugzilla.mozilla.org/show_bug.cgi?id=918281#c15

Thank you for looking into this anyway :-)

PR #413, opened 2015-03-09
Bug 1139301 - Update perfherder date format

tojonmz:
This is supplemental to fixed bug [1139301](https://bugzilla.mozilla.org/show_bug.cgi?id=1139301).

Nothing exciting, just making the perf date format consistent with the rest of treeherder.

Not knowing how else the date  might be reused in perf I chose not to use an angular date filter; additionally, injecting a `thDateFormat` into perf appears problematic because you can't easily have two ng-apps (treeherder, perf) live and be accessed from the same values.js file. So unless ng-app=perf becomes ng-app=treeherder, it's a moot point. I chatted with @camd and he agreed.

Here's the current date style:

perfdateformatcurrent

And proposed (same as elsewhere in Treeherder):

perfdateformatproposed

I'm easy though. If there's compelling UX reasons for keeping the date format as-is for perf, that makes total sense to me.

Tested on OSX 10.9.5:
FF Release **36.0.1**

Adding @wlach  for review and @edmorley for visibility.


tojonmz:

PR #415, opened 2015-03-10
Bug 1137927 - Persist job details panel when jobs are pinned

tojonmz:
This tweak fixes Bugzilla bug [1137927](https://bugzilla.mozilla.org/show_bug.cgi?id=1137927).

In it, the the job panel will persist if the user clicks in an empty main page area, if there are jobs pinned.

The **esc** shortcut and the [x] close icon will continue to work as normal, closing the job panel whether or not jobs are pinned.

There are no visible changes, but here's a screen grab of the scenario:

persistjobpanelpinnedjobs

It's an interesting UX change, I think we want to expose users to it on dev, before pushing it to stage/prod. Without prior knowledge of the change in behavior, it may feel a bit strange for users.

Tested on OSX 10.9.5:
FF Release **36.0.1**

### DIFF
--- a/webapp/app/js/controllers/main.js
+++ b/webapp/app/js/controllers/main.js
@@ -43,7 +43,7 @@ treeherder.controller('MainCtrl', [
             // Suppress for various UI elements so selection is preserved
             var ignoreClear = element.hasAttribute("ignore-job-clear-on-click");
 
-            if (!ignoreClear) {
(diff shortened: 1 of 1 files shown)


tojonmz:
@wlach has r+'d this in the bug, and we'd like to get it evaluated on dev, so I will merge.

PR #416, opened 2015-03-11
Bug 1137456 - Consolidate the pinboard save/clear-all cluster

tojonmz:
This work fixes Bugzilla bug [1137456](https://bugzilla.mozilla.org/show_bug.cgi?id=1137456).

This consolidates the clear-all button as a drop down inside the existing menu. While there I de-cluttered and removed the `fa` icons, and also added in a check so we also expose the drop down menu if there is just a related-bug. This allows them to clear-all via the UI, like they would via the shortcut.

Here's the before:

saveclusterpinboardcurrent

Here's the after (empty pinboard, and populated):

saveclusterpinboardproposed

saveclusterpinboardproposeddropdown

Everything seems to be working fine on Firefox and Chrome.

If we wanted later we could add a `hasComment()` pinboard service api or similar, and use that check also to enable the dropdown if the user types a comment into an otherwise empty pinboard panel. Perhaps we could also disable/dim the 'Save * only' menu entries if no job/bug is present. However neither of those are regressions from the current UI.

Tested on OSX 10.9.5:
FF Release **36.0.1**

### DIFF
--- a/webapp/app/css/treeherder.css
+++ b/webapp/app/css/treeherder.css
@@ -1027,10 +1027,6 @@ ul.failure-summary-list li .btn-xs {
     width: 70px;
 }
 
-#pinboard-controls .clear-all-spacer {
-    width: 3px;
-}
-
 .pinned-job {
     margin-bottom: 2px;
     padding: 1px 2px;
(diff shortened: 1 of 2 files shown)


webapp/app/css/treeherder.css
```diff
@@ -1041,10 +1037,6 @@ ul.failure-summary-list li .btn-xs {
     padding: 1px 2px 1px 2px;
 }
 
-.btn-icon-top-margin {
-    margin-top: 0.3em;
-}
-
```
tojonmz: These two css entries are only used by the save cluster, so are safe for removal as far as I can tell.

tojonmz:
@edmorley has r+'d this in the bug, so I will merge.

PR #417, opened 2015-03-12
Bug 1142559 - Exclude json logs using a blacklist rather than whitelist

edmorley:
Since otherwise we accidentally exclude logs other submitters (eg TaskCluster), since they all have arbitrary names.


maurodoglio:
:shipit:

PR #419, opened 2015-03-12
Bug 1142244 - Improve readability of cancelled job colors

tojonmz:
This tweak fixes Bugzilla bug [1142244](https://bugzilla.mozilla.org/show_bug.cgi?id=1142244).

This improves the readability of the cancelled job color for @AutomatedTester, and more closely matches that of TBPL. Initially we had implemented a paler pink to de-emphasize cancelled jobs. Sheriffs are fine with this change.

We start with TBPL's `#f0c` and reduce the saturation to 75% to match the same sat of blue/green jobs. So we end up at at pink of `#ff40d9`. And for the corresponding roughly 50% saturation field set color, we end up at `#ff80e5`.

Here is the before:

cancelledjobscurrent

And here is proposed:

cancelledjobsproposed

If it's too vibrant we can always tone it down a little, if needed.

Tested on OSX 10.9.5:
FF Release **36.0.1**

### DIFF
--- a/webapp/app/css/treeherder.css
+++ b/webapp/app/css/treeherder.css
@@ -1626,15 +1626,15 @@ fieldset[disabled] .btn-black.active {
 }
 
 .btn-pink {
-  color: rgba(250, 115, 172, 0.63);
+  color: #ff40d9;
   font-weight: bold;
 }
 .btn-pink:hover,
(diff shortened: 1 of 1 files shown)


tojonmz:

PR #420, opened 2015-03-12
Fix bug 1108837 - "Click to add a related bug" text should probably also...

kewisch:
... be clickable.

### DIFF
--- a/webapp/app/css/treeherder.css
+++ b/webapp/app/css/treeherder.css
@@ -972,6 +972,17 @@ ul.failure-summary-list li .btn-xs {
     margin-bottom: -1px;
 }
 
+.pinboard-related-bugs-btn > .pinboard-related-bugs-link,
+.pinboard-related-bugs-btn:hover > .pinboard-related-bugs-link:hover {
+    border-top-right-radius: 3px !important;
+    border-bottom-right-radius: 3px !important;
+}
+
+.pinboard-related-bugs-btn:hover > .pinboard-related-bugs-link {
+    border-top-right-radius: 0 !important;
+    border-bottom-right-radius: 0 !important;
+}
+
 /* Spin button suppression on chrome */
 #pinboard-related-bugs form input[type=number]::-webkit-inner-spin-button {
     -webkit-appearance: none;
@@ -1035,6 +1046,7 @@ ul.failure-summary-list li .btn-xs {
 
 .pinned-job-close-btn {
     padding: 1px 2px 1px 2px;
+    border-color: #fafafa #fafafa #fafafa transparent;
 }
 
 .annotations-panel {
--- a/webapp/app/plugins/pinboard.js
+++ b/webapp/app/plugins/pinboard.js
@@ -5,9 +5,9 @@
 "use strict";
 
 treeherder.controller('PinboardCtrl', [
-    '$scope', '$rootScope', 'thEvents', 'thPinboard', 'thNotify', 'ThLog',
+    '$scope', '$rootScope', '$document', '$timeout','thEvents', 'thPinboard', 'thNotify', 'ThLog',
     function PinboardCtrl(
-        $scope, $rootScope, thEvents, thPinboard, thNotify, ThLog) {
+        $scope, $rootScope, $document, $timeout, thEvents, thPinboard, thNotify, ThLog) {
 
         var $log = new ThLog(this.constructor.name);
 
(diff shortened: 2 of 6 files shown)


webapp/app/css/treeherder.css
```diff
@@ -1035,6 +1046,7 @@ ul.failure-summary-list li .btn-xs {
 
 .pinned-job-close-btn {
     padding: 1px 2px 1px 2px;
+    border-color: #fafafa #fafafa #fafafa transparent;
```
tojonmz: I wonder if these can use shorthand hex, eg. `#fa`.

webapp/app/plugins/pinboard.js
```diff
@@ -108,12 +108,28 @@ treeherder.controller('PinboardCtrl', [
...
+                $scope.newEnteredBugNumber = '';
+            }
         };
 
         $scope.completeClassification = function() {
             $rootScope.$broadcast('blur-this', "classification-comment");
         };
 
+        $scope.checkCancelEnterBugNumber = function(event) {
+            if (event.keyCode == 27 /* DOM_VK_ESCAPE */ || event.key == 'r') {
```
tojonmz: I think we'd like to Mousetrap rather vs. keyCodes if/where possible, for new shortcut functionality. Perhaps this is one of those cases. You can also check controllers/main if of interest, as the core shortcuts are in that file.

webapp/app/css/treeherder.css
kewisch: afaik shorthand hex only works for codes where the R, G or B component has the same letter. So #FFCC00 can be shortened to #fc0, but #fa is not valid.

webapp/app/plugins/pinboard.js
```diff
@@ -108,12 +108,28 @@ treeherder.controller('PinboardCtrl', [
         $scope.toggleEnterBugNumber = function(tf) {
             $scope.enteringBugNumber = tf;
             $scope.focusInput = tf;
+            if (tf) {
+                // We want escape to first cancel entering the bug number, so
+                // unbind the hotkey. Pressing escape again will close the
+                // panels as usual.
+                Mousetrap.unbind(['escape']);
```
tojonmz: I'm pretty sure you don't need the `[]` for a single key, they are used only for lists of keys.

webapp/app/css/treeherder.css
```diff
@@ -972,6 +972,17 @@ ul.failure-summary-list li .btn-xs {
...
+.pinboard-related-bugs-btn > .pinboard-related-bugs-link,
+.pinboard-related-bugs-btn:hover > .pinboard-related-bugs-link:hover {
+    border-top-right-radius: 3px !important;
+    border-bottom-right-radius: 3px !important;
+}
+
+.pinboard-related-bugs-btn:hover > .pinboard-related-bugs-link {
+    border-top-right-radius: 0 !important;
+    border-bottom-right-radius: 0 !important;
+}
```
tojonmz: I'm sure the css is required :) but wondering if you can elaborate on the need for it here and in the markup, and what it does for @camd when he gets to the review.

Wondering/asking if the `!important` is truly needed, just in case it can be omitted.

tojonmz:
Thanks @kewisch ! I made a few triage comments before @camd gets to the review.

tojonmz:
Oh, ya, I'm wondering if on the next commit maybe we could tweak the commit message a bit to describe more what the fix does. ie. sometimes Bugzilla bug summaries are nicely 'declarative', sometimes they are less so :) It's ok if the commit message/PR deviates from the bug summary in that scenario.

eg.
`Bug 1108837 - Support clearing input content in add-related-bugs`

Or whatever message/PR title makes sense to you.

webapp/app/plugins/pinboard.js
kewisch: Maybe I need to read up on Mousetrap, but these are not global hotkeys. They are specific to the related bug input field.

webapp/app/plugins/pinboard.js
kewisch: ok

webapp/app/css/treeherder.css
kewisch: I usually try to avoid `!important`, although I don't know off hand why I used it here. I think there were other css rules with superceeding css specificity and I didn't come up with a readable alternative.

The CSS here is to make the rounded edges look nice when hovering over the related bug input. Steps: enter and accept a related bug. Then hover over the bug number. Before, the right side had hard edges, now it will look nice and rounded. Then hover over the X button, it should now only be rounded on the right of the X button, as before.

kewisch:
I believe the commit message is fine in this case. Its all in one commit, and it actually fixes the case "Click to add a related bug" text should also be clickable. I can remove "Fix" and "probably" to make it more concise.

webapp/app/css/treeherder.css
kewisch: I checked again, this aims to overwrite a bootstrap rule `.btn-group > .btn:first-child:not(:last-child):not(.dropdown-toggle)` which has specifity of 0/0/5/0 while my rule has 0/0/2/0. I don't think it makes sense to add in an id selector, nor add 4 class/attribute selectors.

kewisch:
Everything updated. I couldn't use the simple blur handler due to timing issues, but this works. Also a fix to use ng-keydown instead of ng-keypress so it also works on chrome.

webapp/app/plugins/pinboard.js
camd: This seems to work well, but I have to agree with @tojonmz.  Can we make this work with Mousetrap for consistency.  We're trying to get away from using ``keyCode``.

kewisch: Ok, I'll look into this.

webapp/app/css/treeherder.css
tojonmz: Ah, ya sorry, my bad. :)

webapp/app/plugins/pinboard.js
kewisch: Taken care of, it seems rebinding with Mousetrap works without issues. I checked with a debugger breakpoint to make sure its not still bound to closing the bug field after the pinboard is closed.

camd: @kewisch It looks the same to me.  Did your changes get pushed up to origin yet?

kewisch: Oops, sorry about that. git commit --amend fooled me because I forgot git add -u. Pushed correctly now.

camd:
Looks great!  Thanks!!

PR #424, opened 2015-03-16
Bug 1139599 - Option to highlight revision in perfherder

mishravikas:


webapp/app/js/perf.js
```diff
@@ -310,6 +334,40 @@ perf.controller('PerfCtrl', [ '$state', '$stateParams', '$scope', '$rootScope',
...
+      $scope.resetHighlightButton = false;
+      updateURL();
+      plotGraph();
+    }
+
+    $scope.highlightRevision = function() {
+      var rev = $scope.revisionToHighlight;
+      if (rev.length == 12) {
+        var plotdata = $scope.plot.getData();
+        plotdata.forEach(function(series, i) {
```
wlach: Instead of doing this several times (one for each series), why don't we call map on a set of $http promises, and then call plotGraph once at the end. You can see an example of this here: https://github.com/mozilla/treeherder-ui/blob/master/webapp/app/js/perf.js#L238 Also, unless I'm missing something, we should be able to call updateURL just once, at the beginning. 

i.e. 
    updateURL();
    $q.all($scope.plot.getData().map(function(series) { 
      if (series.points.show) { return $http.get(...).then(function(response) { ... }); }
      return null; 
     }));

webapp/app/js/perf.js
```diff
@@ -310,6 +336,48 @@ perf.controller('PerfCtrl', [ '$state', '$stateParams', '$scope', '$rootScope',
...
+             "/resultset/?format=json&revision=" + rev + "&with_jobs=false").then(
+            function(response) {
+              if (response.data.results.length > 0) {
+                var result_set_id = response.data.results[0].id;
+                var j = series.flotSeries.resultSetData.indexOf(result_set_id);
+                var seriesToaddHighlight = _.find($scope.seriesList, function(sr) { return sr.signature == series.signature });
+                seriesToaddHighlight.highlighted = [j, rev];
+              }
+              updateURL();
+              plotGraph();
```
wlach: We should do this when the q.all() promise is finished, rather than every time we download from http.

wlach:
This is getting a lot closer. A few edge cases still don't work right here (in my brief testing):

1. If I add a new series, the relevant revision isn't highlighted automatically.
2. If I delete the contents of the textbox, the "x" is still present.

webapp/app/css/perf.css
```diff
@@ -150,3 +150,9 @@ to { opacity: 0; }
   overflow-y: auto;
 }
 
+.resetHighlightButton {
```
wlach: to be consistent with other css directives, call this .reset-highlight-button

webapp/app/js/perf.js
```diff
@@ -164,11 +166,31 @@ perf.controller('PerfCtrl', [ '$state', '$stateParams', '$scope', '$rootScope',
...
+      });
+    }
+
     function plotGraph() {
       // synchronize series visibility with flot, in case it's changed
       $scope.seriesList.forEach(function(series) {
         series.flotSeries.points.show = series.visible;
       });
+
+      // Show graph-bottom if there's atleast one series
```
wlach: This code isn't necessary. You can just evaluate whether seriesList.length is greater than 0 directly in the template (see below)

webapp/app/partials/perf/perfctrl.html
```diff
@@ -37,6 +37,10 @@
     </select>
     <hr/>
     <div id="graph"></div>
+    <div id="graph-bottom" ng-show="showGraphBottom">
```
wlach: To remove the need for the showGraphBottom function, just put the expression seriesList.length > 0 in ng-show instead.

webapp/app/partials/perf/perfctrl.html
```diff
@@ -37,6 +37,10 @@
...
+        Highlight revision: <input type="text" ng-change="addHighlightedRevision()" placeholder="12 character hg revision" ng-model="revisionToHighlight">
+        <div class="resetHighlightButton" ng-show="resetHighlightButton" ng-click="resetHighlight()">&#10006;</div>
```
wlach: Likewise, you should be able to replace the resetHighlightButton variable with some kind of expression here. Don't know what it is off-hand but I'll leave it to you to figure out. :)

wlach:
Merged manually

PR #425, opened 2015-03-17
Bug 1143666 - Add a shortcut for retrigger job

tojonmz:
This fixes Bugzilla bug [1143666](https://bugzilla.mozilla.org/show_bug.cgi?id=1143666).

This adds the '**r**' shortcut for retriggering a selected job. It also re-maps the add-related-bugs shortcut, which was '**r**', to '**b**'.

Everything seems to be working ok locally, but for a variety of unrelated reasons (local vagrant issue, LDAP access) I can't actually issue a retrigger. I just used a console print to make sure the shortcut was getting there with no lag, and it also correctly gave me the log-in warning balloon if I wasn't logged in.

retriggerscenario

The add-related-bugs shortcut seems to continue to work fine using its new '**b**', and integration testing with other pinboard shortcuts and next/prev job navigation, seems fine.

If you can try my branch and actually retrigger a job or two with your own LDAP access using the shortcut, that would be awesome. :)

(sort of) Tested on OSX 10.9.5:
FF Release **36.0.1**

Adding @edmorley for review.

### DIFF
--- a/webapp/app/js/controllers/main.js
+++ b/webapp/app/js/controllers/main.js
@@ -56,9 +56,10 @@ treeherder.controller('MainCtrl', [
                 'n',     // Select next unclassified failure
                 'k',     // Select previous unclassified failure
                 'p',     // Select previous unclassified failure
+                'r',     // Retrigger selected job
                 'space', // Pin selected job to pinboard
                 'u',     // Display only unclassified failures
-                'r',     // Pin selected job and add related bug
+                'b',     // Pin selected job and add related bug
                 'c',     // Pin selected job and add classification
                 'f',     // Enter a custom job or platform filter
                 'left',  // Select previous job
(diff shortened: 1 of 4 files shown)


webapp/app/js/controllers/main.js
```diff
@@ -109,6 +110,13 @@ treeherder.controller('MainCtrl', [
                 $rootScope.$emit(thEvents.selectPreviousUnclassifiedFailure);
             });
 
+            // Shortcut: retrigger selected job
+            Mousetrap.bind('r', function() {
+                $scope.$evalAsync(
+                    $rootScope.$emit(thEvents.jobRetrigger, $rootScope.selectedJob)
```
edmorley: For the 'space' and 'c' shortcuts, we check to see if a job is selected _before_ triggering the event, rather than on the listeners side. If there's a case for doing the latter then I'm all ears - but it seems like this should be doing the former otherwise, for consistencies sake? :-)

tojonmz: Totally correct. Ed you are awesome, thanks for catching it. We should always test for selected job at shortcut invocation when applicable. I've switched it and it seems to be working fine in Firefox and Chrome and with other shortcut integration testing (related bugs, comment, save, clear, pin). I'll push the delta, then squash after you've had a look.

edmorley:
lgtm now :-)

PR #426, opened 2015-03-18
Bug 1143798 - Show visual indication of retrigger for non bb jobs

lightsofapollo:


webapp/app/js/services/buildapi.js
```diff
@@ -32,7 +32,7 @@ treeherder.factory('thBuildApi', [
     return {
         retriggerJob: function(repoName, requestId) {
 
-            $http({
+            return $http({
```
lightsofapollo: tldr; Return promise so UI instead of model can decide what error message to show.

PR #427, opened 2015-03-18
Bug 1144748 - Improve readability of help shortcuts with keyboard icons

tojonmz:
This update fixes Bugzilla bug [1144748](https://bugzilla.mozilla.org/show_bug.cgi?id=1144748).

In it we improve the readability of the shortcuts in Help. I just created a non-specific `kbd` class which we could use anywhere, but referenced its current use in treeherder.css.

Here's the before:

shortcutshelpcurrent

Here's the after:

shortcutshelpproposed

Everything seems fine on both browsers.

Tested on OSX 10.9.5:
FF Release **36.0.1**

Adding @edmorley for review.

### DIFF
--- a/webapp/app/css/treeherder.css
+++ b/webapp/app/css/treeherder.css
@@ -1724,6 +1724,20 @@ fieldset[disabled] .btn-repo.active {
     overflow: auto;
 }
 
+/* Used in Help but for general use */
+.kbd {
+    display: inline-block;
+    padding: 3px 5px;
+    font: 11px/10px Consolas,"Liberation Mono",Menlo,Courier,monospace;
+    color: #555;
+    vertical-align: middle;
+    background-color: #fcfcfc;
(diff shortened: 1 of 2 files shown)


tojonmz:
Actually I suppose I could use the `<kbd>` tag instead of a class.

webapp/app/css/treeherder.css
```diff
@@ -1724,6 +1724,20 @@ fieldset[disabled] .btn-repo.active {
     overflow: auto;
 }
 
+/* Used in Help but for general use */
+.kbd {
+    display: inline-block;
+    padding: 3px 5px;
+    font: 11px/10px Consolas,"Liberation Mono",Menlo,Courier,monospace;
```
tojonmz: Using Github's own font family order here.

tojonmz:
We decided just to stay with spans for now.

PR #429, opened 2015-03-19
Bug 1145281 - Enable manual entry in highlight revision box

mishravikas:


wlach:
Looks good. While you're here, could you specify a maxlength="12" property on the input element, so people don't accidentally put in more than 12 characters?

PR #430, opened 2015-03-19
Bug 1133021 - Linkify all bugs and prs in revision commit messages

tojonmz:
This work fixes Bugzilla bug [1133021](https://bugzilla.mozilla.org/show_bug.cgi?id=1133021).

This linkifies all bugs and pull requests in a given commit message that match the following prefixes:

* "Bug ", "bug ", "-- "
* "PR#" "pr#"

Here's the before with an example test case (note also the commit-title populating the bug/pr links):

commitlinkscurrent

Here's the after:

commitlinksproposed

I've tested a variety of different messages (no bugs, single bugs/prs, multiple bugs/prs, identical bugs/prs, identically-prefixed bugs/prs, the originally reported bug revision ) and everything seems to be ok.

I tried a bunch of different implementations: loops, a function with passed params, doing the replace in one shot (problematic with "-- " and the $n param in the url vars), specifying the .replace string as a [function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_function_as_a_parameter) which also doesn't seem compatible with $n use.

So instead we just check if we have pr or bug match, and do a separate pass for each case. We incur minor cost, but probably no more than if we tested for conditions.

We now inject a generic bmo/github title for each issue type. The overall span retains its own tooltip commit summary on hover. Initially I did it to avoid a bmo/github API to get the "Summary", which was outside the scope of the bug anyway. I think a generic title makes more sense. It tells the user what the link is, and doesn't confuse them with what could be a very similar tooltip to the commit span.

I've left some test values in this initial commit for testing, which I will remove prior to merge. Everything appears to be ok on both Firefox and Chrome. Please beat it up as much as possible and with other commit message sequences :)

Tested on OSX 10.9.5:
FF Release **36.0.1**

Adding @wlach for review/branch testing, and @edmorley for visibility.

### DIFF
--- a/webapp/app/js/filters.js
+++ b/webapp/app/js/filters.js
@@ -40,22 +40,32 @@ treeherder.filter('stripHtml', function() {
 treeherder.filter('linkifyBugs', function() {
     return function(input) {
         var str = input || '';
-        var bug_matches = /.*-- ([0-9]+)|.*bug-([0-9]+)|.*Bug ([0-9]+)/ig.exec(str);
-        var pr_matches = /PR#([0-9]+)/i.exec(str);
         var clear_attr = 'ignore-job-clear-on-click';
-        if (pr_matches) {
-            var pr_url = "https://github.com/mozilla-b2g/gaia/pull/" + pr_matches[1];
-            var pr_hyperlink = '<a href="' + pr_url + '" ' + clear_attr + '>' +
-                               pr_matches[1] + '</a>';
-            str = str.replace(pr_matches[1], pr_hyperlink);
-        }
+
+        var bug_matches = str.match(/-- ([0-9]+)|bug.([0-9]+)/ig);
+        var pr_matches = str.match(/PR#([0-9]+)/ig);
+
+        // Settings
+        var bug_title = 'bugzilla.mozilla.org';
+        var bug_url = '<a href="https://bugzilla.mozilla.org/show_bug.cgi?id=$1" ' +
+            'data-bugid=$1 ' + 'title=' + bug_title + '>$1</a>';
+        var pr_title = 'github.com';
+        var pr_url = '<a href="https://github.com/mozilla-b2g/gaia/pull/$1" ' +
+            'data-prid=$1 ' + 'title=' + pr_title + '>$1</a>';
+
         if (bug_matches) {
-            var bug_match = bug_matches[1] || bug_matches[2] || bug_matches[3];
-            var bug_url = "https://bugzilla.mozilla.org/show_bug.cgi?id=" + bug_match;
-            var bug_hyperlink = '<a href="' + bug_url + '" ' + clear_attr + '>' +
-                                bug_match + '</a>';
-            str = str.replace(bug_match, bug_hyperlink);
+            // Separate passes to preserve prefix
+            str = str.replace(/Bug ([0-9]+)/g, "Bug " + bug_url);
+            str = str.replace(/bug ([0-9]+)/g, "bug " + bug_url);
+            str = str.replace(/-- ([0-9]+)/g, "-- " + bug_url);
         }
+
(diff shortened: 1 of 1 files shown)


webapp/app/js/filters.js
```diff
@@ -40,22 +40,39 @@ treeherder.filter('stripHtml', function() {
...
-                               pr_matches[1] + '</a>';
-            str = str.replace(pr_matches[1], pr_hyperlink);
-        }
+
+        // Test strings
+        // var str = "No bugs";
+        // var str = "One bug 1130488";
+        // var str = "One PR#456";
+        // var str = "Bug 1130488 disable tests for bug 999 and -- 777 and PR#368 and PR#456";
+        var str = "Bug 1130488 Bug 1130488 bug 1130488 -- 777 and pr#368 and PR#456";
```
wlach: I don't think you meant to leave this in. :)

webapp/app/js/filters.js
```diff
@@ -40,22 +40,39 @@ treeherder.filter('stripHtml', function() {
...
         if (bug_matches) {
-            var bug_match = bug_matches[1] || bug_matches[2] || bug_matches[3];
-            var bug_url = "https://bugzilla.mozilla.org/show_bug.cgi?id=" + bug_match;
-            var bug_hyperlink = '<a href="' + bug_url + '" ' + clear_attr + '>' +
-                                bug_match + '</a>';
-            str = str.replace(bug_match, bug_hyperlink);
+            // Separate passes to preserve prefix
+            var str = str.replace(/Bug ([0-9]+)/g, "Bug " + bug_url);
+            var str = str.replace(/bug ([0-9]+)/g, "bug " + bug_url);
+            var str = str.replace(/-- ([0-9]+)/g, "-- " + bug_url);
```
wlach: Don't redeclare "var str" here, it's already declared above. Just say str = ...

webapp/app/js/filters.js
```diff
@@ -40,22 +40,39 @@ treeherder.filter('stripHtml', function() {
...
         }
+
+        if (pr_matches) {
+            // Separate passes to preserve prefix
+            var str = str.replace(/PR#([0-9]+)/g, "PR#" + pr_url);
+            var str = str.replace(/pr#([0-9]+)/g, "pr#" + pr_url);
```
wlach: As above, no need to redeclare str

webapp/app/js/filters.js
```diff
@@ -40,22 +40,39 @@ treeherder.filter('stripHtml', function() {
...
-            var pr_hyperlink = '<a href="' + pr_url + '" ' + clear_attr + '>' +
-                               pr_matches[1] + '</a>';
-            str = str.replace(pr_matches[1], pr_hyperlink);
-        }
+
+        // Test strings
+        // var str = "No bugs";
+        // var str = "One bug 1130488";
+        // var str = "One PR#456";
+        // var str = "Bug 1130488 disable tests for bug 999 and -- 777 and PR#368 and PR#456";
```
wlach: You can take these comments out, there's no need for them in production code as long as it works. :)

webapp/app/js/filters.js
tojonmz: Gah, yes. Of course.

webapp/app/js/filters.js
tojonmz: Actually was just for your local testing (per PR), along with the others :) But ya, if you can just try it locally with your own flavors that would be awesome. I will tweak and re-push.

wlach:
This looks fine to me. :) Feel free to merge if it's working well.

tojonmz:
Yup, it seems to be working fine in more local testing. I will merge.

PR #431, opened 2015-03-20
Bug 1145311 - Support y-axis zoomng in perfherder

mishravikas:


jmaher:
this is pretty straightforward.  I am not 100% familiar with this code.  My understanding of these changes are they are safe; Any unknowns would be on areas of the code which I had overlooked.

PR #434, opened 2015-03-20
Bug 1082771 - Add a static "Infrastructure" menu that links to existing dashboards

edmorley:
Suggestions welcome for additions/removals/re-ordering of links.
@tojonmz I'm sure you'll have some styling ideas for this - I don't know whether it's best to land as is - and then let you do your (awesome) thing - or iterate here? I don't mind either way - though with the PTO might be hard to get this landed before TBPL end of life if we aim for styling perfection here :-)

screenshot

### DIFF
--- a/webapp/app/partials/main/thGlobalTopNavPanel.html
+++ b/webapp/app/partials/main/thGlobalTopNavPanel.html
@@ -19,6 +19,18 @@
                 </span>
             </span>
 
+            <!-- Infra Menu -->
+            <span class="repo-menu dropdown">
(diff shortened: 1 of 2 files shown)


tojonmz:
@edmorley this looks good (or without the dividers per the IRC thread a few mins ago), I'd say we go as-is and any tweaking we can do later.

The only conspicuous thing all these dropdowns across the app of this class seems the font-size, which seems a bit big (eg. also the pinboard Save button dropdown), but is independent of your awesome work above :)

PR #437, opened 2015-03-21
Bug 1112652 -  Communicate the resultset date link is a permalink/url

tojonmz:
This tweak hopefully addresses the bulk of Bugzilla bug [1112652](https://bugzilla.mozilla.org/show_bug.cgi?id=1112652), or enough we address the use problem of discoverability of a sharable url, for a single resultset.

Here's the before:

screen shot 2015-03-21 at 2 08 54 pm

Here's the after:

screen shot 2015-03-21 at 12 53 30 pm

I went through a variety of other wordings, some which I've included in case folks like anything better:

* "View only this resultset and url"
* "View only this resultset url"
* "View this resultset url"

Feel free to let me know if prefer any of those, or propose others, and I'll adjust :)

Tested on OSX 10.9.5:
FF Release **36.0.1**

Adding @wlach for review and @armenzg and @KWierso for visibility.

### DIFF
--- a/webapp/app/css/treeherder.css
+++ b/webapp/app/css/treeherder.css
@@ -376,6 +376,12 @@ th-watched-repo {
 
 }
 
+.result-set-superscript {
+    vertical-align: super;
+    font-size: 0.625em;
+    margin-left: -0.2em;
+}
+
 .result-set {
     padding-bottom: 0px;
     margin-left: 0;
@@ -397,7 +403,7 @@ th-watched-repo {
 }
 
 .result-set-title-left {
-    flex: 0 0 23.5em;
+    flex: 0 0 24.2em;
     overflow: hidden;
     text-overflow: ellipsis;
     padding-right: 10px;
--- a/webapp/app/partials/main/jobs.html
+++ b/webapp/app/partials/main/jobs.html
@@ -14,8 +14,9 @@
           <span class="btn btn-default btn-sm glyphicon glyphicon-download" ng-show="isLoadingJobs"></span>
           <span>
             <a href="{{::revisionResultsetFilterUrl}}"
-               title="open this resultset"
-               ignore-job-clear-on-click>{{::resultsetDateStr}}</a> - </span>
+               title="View only this resultset"
+               ignore-job-clear-on-click>{{::resultsetDateStr}}
+              <span class="fa fa-external-link result-set-superscript"></span></a> - </span>
           <th-author author="{{::resultset.author}}"></th-author>
         </span>
         <span class="revision-text">{{::resultset.revision}}</span>


KWierso:
I would argue that the tooltip alone isn't enough to communicate this to the user. The permalink icon would, though.

tojonmz:
What if we replaced the '-' character between the date and author, with the `fa-link`. At least that way we'd not be adding clutter, and we be replacing an existing separator with something which communicates the functionality. Since the values on both sides are links.

wlach:
Maybe just something like "View only this resultset"? I find the URL terminology a bit confusing

tojonmz:
Sure, I will try that wording. I also think Wes's idea for icon left-side makes more sense now that I think about it. ie. `[icon] Date [-] Author`

Since 'Author' produces dynamic results, and is not perma. I'll mock up both locations and put them up for you guys to look at.

tojonmz:
Here's `fa-link` at the beginning only:

fa-link-span

Here's the same but as a replacement for the "-" separator:

fa-link-span-middle-dash-replace

Here's `fa-external-link`:

fa-external-link-span

And the same but as a part of the anchor:

fa-external-link-linked

The last one still has some font-weight change to sort out. Also the vertical alignment needs to be sorted on all of them. But a selection to spark ideas. :)

KWierso:
I like 1 and 4. :) 

Does 1 turn blue as part of the anchor, too?

tojonmz:
I made the grey ones just 'inert' spans for the screen grabs, they aren't part of the anchor. But sure we could make any of them part of the anchor and always be blue (like 4).

I agree I kind of like 4 `fa-external-link` also, since we already use a solid flavor of it at the bottom of our resultset commit list "and more...".

We could also apply a rotate to `fa-link` 1), eg. 45 deg or -45 deg and make the chain totally vertical or horizontal if we wanted.

The fa-link in 1) seems.. err.. a lot like a piece of text in its line weight when you blur your eyes. Whereas 4) seems quite open and differentiated from the adjacent text. Just my take though.

tojonmz:
Here's 1 and 4 for rough comparison, as part of the anchor:

fa-link-linked

fa-external-link-linked

tojonmz:
> Maybe just something like "View only this resultset"? I find the URL terminology a bit confusing

I wonder about:

* "View this resultset link"
* "View this resultset address"
* "View as a sharable link"

...basically something which addresses @armenzg's initial issue that we aren't communicating it is a static link which can be shared.

Not to bikeshed :) but my take is adding an fa link icon in the UI date string, doesn't really make its purpose discoverable by itself.

tojonmz:
I thought of these while out on a walk. I kinda like them also. Superscript annotations (either grey or blue). They don't disturb the line as much, and don't read as an in-line text character like all the other concepts.

superscriptannotationgrey

superscriptannotationblue

tojonmz:
Chatted with @KWierso on IRC, we both think the above CS mock is pretty groovey, so I did it for real. Here's the current production, proposed, and proposed w/title:

currentproduction

superscriptproposed

superscriptwithhover

tojonmz:
@wlach I think we are good for another review now.

wlach:
Perfect, thank you!

PR #438, opened 2015-03-23
Bug 1135862 - Allow classification dropdown save

tojonmz:
This work fixes Bugzilla bug [1135862](https://bugzilla.mozilla.org/show_bug.cgi?id=1135862).

More LHF, this basically allows the user to '**ctrl**+**enter**' to save a classification if the classification dropdown is focused.

Here's the new workflow showing the save now getting triggered in that state:

saveclassificationdropdownfocusproposed

I also tried to support shortcuts '**c**' and '**b**' (add comment, add related bug) when the classification dropdown is focused, so the user can get at that functionality. There is some peculiarity where it only works on first invocation:

* pin a job
* select a classification type
* '**c**' and enter a comment (works)
* select another classification type
* '**c**' to edit the comment again (doesn't work)

I must be missing something obvious.

All other integration testing, other shortcut interaction, pinboard clearing, seems fine. I changed the code slightly avoid replicating the `allowKeys()` key list.

Tested on OSX 10.9.5:
FF Release **36.0.4**
Chrome Latest Release 41.0.2272.101 (64-bit)

Adding @camd for review.

### DIFF
--- a/webapp/app/js/controllers/main.js
+++ b/webapp/app/js/controllers/main.js
@@ -48,23 +48,33 @@ treeherder.controller('MainCtrl', [
             }
         };
 
-        // Disable single key shortcuts in specified shortcut events
+        // Single key shortcuts to allow in ui events (usually inputs)
+        var mousetrapExclusions = [
+            'i',     // Toggle display in-progress jobs (pending/running)
+            'j',     // Select next unclassified failure
+            'n',     // Select next unclassified failure
+            'k',     // Select previous unclassified failure
+            'p',     // Select previous unclassified failure
+            'r',     // Retrigger selected job
+            'space', // Pin selected job to pinboard
+            'u',     // Display only unclassified failures
+            'b',     // Pin selected job and add related bug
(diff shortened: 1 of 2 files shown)


camd:
I think the "c" not working on the second time has to do with the classification drop down being focused.  It's taking the key and trying to select the list item that starts with that.  So, if I hit "e" it will set it to "expected fail".  Hard to say what the right thing is there.  But hitting tab from that spot will highlight the comment field.  Perhaps it's good as it is.  I'm cool with that if you are.  :)

tojonmz:
Cool thanks @camd. We discussed this >2x 'b' and 'c' shortcut behavior over IRC and we've resolved merge it as-is for now. I'll open a low priority follow up bug for it, in case we can sort out the cause.

tojonmz:
Merged and pushed manually as https://github.com/mozilla/treeherder-ui/commit/e35c2e82dd24a010b80e007c1c80929947d84fc7.

PR #441, opened 2015-03-25
Bug 1139421 - Allow hover copy for Job name

tojonmz:
This work fixes Bugzilla bug [1139421](https://bugzilla.mozilla.org/show_bug.cgi?id=1139421).

In it we allow for hover-copy of the job name, like we do with the logviewer url and raw log url. Instead we provide kind of reverse functionality here, allowing the user to copy raw text from a link. This bypasses the need to drag-select the value.

Here's the scenario, hovering over any Job name:

hoverjobname

And you **ctrl/cmd**+**c**

And paste the text wherever you like:

hovercopypastevalue

I also updated the Help, since I think we are at a point where we need to break out this functionality into a separate section:

hovercopyhelpupdate

Makes it a bit more prominent and discoverable.

I suppose we could make all the `<label>`'s in the Job details pane the links, and make all our current links, plain text. This would have facilitated copying the traditional way. But having linked labels seems kind of strange.. it would seem more appropriate to link labels to a glossary of terms, than a function which does something.

Everything seems to be working as expected on both browsers, switching between all three existing hover copy elements, pasting them, switching between them, filtering by Job name and copying, and switching jobs and copying new values.

Not sure, but think I will need to include the new dark logviewer svg added here, in a Grunt build file or some kind before I merge.

Tested on OSX 10.9.5:
FF Release **36.0.4**

Adding @camd for review and @jmaher and @armenzg for visibility.


webapp/app/css/treeherder.css
```diff
@@ -1808,6 +1808,20 @@ fieldset[disabled] .btn-repo.active {
...
+#help-raw-log-icon {
+    font-size: 13px;
+    color: #9fa3a5;
+}
+
+#help-logviewer-icon {
+    vertical-align: text-bottom;
+    width: 15px;
+}
+
```
tojonmz: Re-using the icons in Help with their existing treeherder classes was a bit problematic, so these ids were added.

camd:
@tojonmz This looks great to me.  Did you say you may make an icon adjustment before merging?  Are you able to merge this yourself now?  Or would you like me to do it?

Nice!  this will be great.

tojonmz:
@camd cool, ya I just thought the new svg needed to be added to some other manifest file in the repo, but I will check tomorrow before I merge.

tojonmz:
I checked with Mauro also this morning, ya I think we are fine. I rebased so my merge-commit will be paired nicely in the commit history, so I will merge now.

PR #442, opened 2015-03-26
Bug 1113322 - tier 2 jobs visual representation

camd:
``Tier-2`` jobs are represented in a group by that name in the UI:

screenshot 2015-03-27 12 38 54

### DIFF
--- a/webapp/app/js/services/jobfilters.js
+++ b/webapp/app/js/services/jobfilters.js
@@ -96,6 +96,10 @@ treeherder.factory('thJobFilters', [
             name: "platform",
             matchType: MATCH_TYPE.substr
         },
+        tier: {
+            name: "tier",
+            matchType: MATCH_TYPE.exactstr
+        },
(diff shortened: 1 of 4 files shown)


webapp/app/js/services/jobfilters.js
```diff
@@ -471,7 +484,7 @@ treeherder.factory('thJobFilters', [
             }
             result = platform + " " + job.platform_option;
         }
-        return result;
+        return String(result);
```
maurodoglio: why are you doing this string conversion?

maurodoglio:
Other that the question above, everything looks good!

webapp/app/js/services/jobfilters.js
camd: Before we were only filtering on fields that were strings in the ``job`` object already.  But with ``tier`` it's an int.  So, I need to cast the ``job.tier`` field value to a string for compare.

PR #443, opened 2015-03-27
Bug 1131830 - Prevent accidental logouts with a dropdown logout menu

tojonmz:
This work fixes Bugzilla bug [1131830](https://bugzilla.mozilla.org/show_bug.cgi?id=1131830).

Per the bug thread with @edmorley and @KWierso, we now prevent the user from accidentally logging themselves out when working quickly near the Logout button by creating a nested menu. While there I added a bit of polish to cue the user they are logged in with Persona and also not occupy so much space. The same information is available to the user as before.

This Persona icon approach will also work nicely with an adjacent 'sheriff on duty' icon (bug [1144589](https://bug1144589.bugzilla.mozilla.org/attachment.cgi?id=8582586)), if we fix that issue also.

Here's the current appearance (logged in):

logoutcurrent

Here's proposed (logged in):

logoutproposed

And on hover:

logoutproposedtitle

And posted menu:

logoutproposedmenu

All the dropdown menus are a bit big, that will be addressed as part of bug [1145837](https://bugzilla.mozilla.org/show_bug.cgi?id=1145837).

I intentionally added an 'Actions' divider in the menu even though we have one entry at present, given we will likely have a 'Settings' divider below it, for example for `.is_staff` users who wish to set themselves as the on-duty sheriff, or other personalized user settings.

Everything seems to be behaving fine on both browsers, and with basic integration testing.

Tested on OSX 10.9.5:
FF Release **36.0.4**

Adding @wlach for review

### DIFF
--- a/webapp/app/partials/main/persona_buttons.html
+++ b/webapp/app/partials/main/persona_buttons.html
@@ -1,7 +1,17 @@
-<a class="btn btn-view-nav btn-right-navbar nav-login-btn"
-   ng-if="user.loggedin"
-   ng-click="logout()">
-   <span>Logout</span></a>
+<span class="dropdown"
+      ng-if="user.loggedin">
+  <button id="logoutLabel" title="Logged in as: {{user.email}}" role="button"
+          href="#" data-toggle="dropdown" data-target="#"
+          class="btn btn-view-nav btn-right-navbar nav-repo-btn">
+    <div class="nav-user-icon">
+      <span class="fa fa-user pull-left"></span>
+    </div>
+    <span class="fa fa-angle-down lightgray"></span>
+  </button>
+  <ul class="dropdown-menu" role="menu" aria-labelledby="logoutLabel"
+      ng-include="'partials/main/thLogoutMenu.html'">
+  </ul>
+</span>
 
 <a class="btn btn-view-nav btn-right-navbar nav-login-btn"
    title="Log in for additional functionality"
--- a/webapp/app/partials/main/thGlobalTopNavPanel.html
+++ b/webapp/app/partials/main/thGlobalTopNavPanel.html
@@ -21,7 +21,7 @@
 
             <!-- Infra Menu -->
             <span class="repo-menu dropdown">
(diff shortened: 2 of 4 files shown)


webapp/app/partials/main/persona_buttons.html
```diff
@@ -1,7 +1,17 @@
-<a class="btn btn-view-nav btn-right-navbar nav-login-btn"
-   ng-if="user.loggedin"
-   ng-click="logout()">
-   <span>Logout</span></a>
+<span class="dropdown"
+      ng-if="user.loggedin">
+  <button id="logoutLabel" title="Logged in as: {{user.email}}" role="button"
+          href="#" data-toggle="dropdown" data-target="#"
+          class="btn btn-view-nav btn-right-navbar nav-repo-btn">
```
tojonmz: nb. we've also used `nav-repo-btn` properties elsewhere eg. for the Infra menu. I plan to sort this out for all menus in a follow up bug.

webapp/app/partials/main/thGlobalTopNavPanel.html
```diff
@@ -21,7 +21,7 @@
 
             <!-- Infra Menu -->
             <span class="repo-menu dropdown">
-                <button id="infraLabel" title="Infrastructure Status" role="button"
+                <button id="infraLabel" title="Infrastructure status" role="button"
```
tojonmz: Unrelated tweak for tooltip Case consistency.

webapp/app/partials/main/thLogoutMenu.html
```diff
@@ -0,0 +1,4 @@
+<li role="presentation" class="dropdown-header">Actions</li>
```
wlach: I'd kill this line for now. There is only one action, no need to label it.

tojonmz:
So the only tweaks we did from the original screen grabs above was to remove the 'Actions' divider in the menu until we have additional entries, and I also tweaked the nav-user-icon down by 1 pixel from 13 to 12px last night as I noticed the navbar height was getting nudged larger on login. Now that doesn't occur. So I will merge.

PR #448, opened 2015-04-01
Bug 1149576 -  Adjust the max-pin count error message for clarity

tojonmz:
This fixes Bugzilla bug [1149576](https://bugzilla.mozilla.org/show_bug.cgi?id=1149576).

This adjusts the max pin count message, so when the user hits the limit on a re-pin, we no longer tell them about how many we added or not. Example workflow:

* pin All
* manually un-pin 1 job
* pin All again

Here's the current appearance:

current

Here's the proposed:

proposed

I considered injecting thPinboard into the various job and filter controllers, but instead decided to pass the maxSize as a new param, and rename the old max size as the `spaceRemaining` param, since that is what it is, upstream.

I thought this might be more economical than more injection. Everything seems fine in the Resultset pin-all, and the Filter Panel pin-all workflows.

`getAllShownJobs` seems to be only used with `pinAllShownJobs` so I am assuming the referencing "pinboard size" in the message will be fine for now.

Tested on OSX 10.9.5:
FF Release **36.0.4**

Adding @camd for review.

### DIFF
--- a/webapp/app/js/controllers/filters.js
+++ b/webapp/app/js/controllers/filters.js
@@ -161,7 +161,8 @@ treeherder.controller('FilterPanelCtrl', [
             }
             var shownJobs = ThResultSetStore.getAllShownJobs(
                 $rootScope.repoName,
-                thPinboard.spaceRemaining()
+                thPinboard.spaceRemaining(),
+                thPinboard.maxNumPinned
             );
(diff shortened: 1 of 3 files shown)


tojonmz:
@camd no issues with the two commits above (identical code), I was just monkeying around with a rebase to test something.

PR #450, opened 2015-04-03
Fix ``job`` field value for Task Cluster jobs and allow filtering

camd:
Please merge this service PR first: https://github.com/mozilla/treeherder-service/pull/447

This fixes:
* ``job`` field in the details panel for Task Cluster jobs will be accurate and not refer to last Buildbot job that was selected.
* You can filter the Task Cluster job just like you can with a Buildbot buildername.
* Adds the job ``signature`` field to the search string to make it more accurate (we were often getting many other jobs when clicking this link to filter on Buildbot jobs).

### DIFF
--- a/webapp/app/js/directives/clonejobs.js
+++ b/webapp/app/js/directives/clonejobs.js
@@ -9,12 +9,14 @@ treeherder.directive('thCloneJobs', [
     '$rootScope', '$http', 'ThLog', 'thUrl', 'thCloneHtml',
     'thServiceDomain', 'thResultStatusInfo', 'thEvents', 'thAggregateIds',
     'thJobFilters', 'thResultStatusObject', 'ThResultSetStore',
-    'ThJobModel', 'linkifyBugsFilter', 'thResultStatus', 'thPlatformNameMap',
+    'ThJobModel', 'linkifyBugsFilter', 'thResultStatus', 'thPlatformName',
+    'thJobSearchStr',
     function(
         $rootScope, $http, ThLog, thUrl, thCloneHtml,
         thServiceDomain, thResultStatusInfo, thEvents, thAggregateIds,
         thJobFilters, thResultStatusObject, ThResultSetStore,
-        ThJobModel, linkifyBugsFilter, thResultStatus, thPlatformNameMap){
+        ThJobModel, linkifyBugsFilter, thResultStatus, thPlatformName,
+        thJobSearchStr){
 
     var $log = new ThLog("thCloneJobs");
 
@@ -97,11 +99,11 @@ treeherder.directive('thCloneJobs', [
         if (direction === 'next') {
             getIndex = function(idx, jobs) {
                 return idx+1 > _.size(jobs)-1 ? 0: idx+1;
-            }
+            };
         } else if (direction === 'previous') {
             getIndex = function(idx, jobs) {
                 return idx-1 < 0 ? _.size(jobs)-1 : idx-1;
-            }
+            };
         }
 
         jobs = $(".th-view-content .job-btn");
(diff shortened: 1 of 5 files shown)


webapp/app/plugins/controller.js
```diff
@@ -108,9 +110,15 @@ treeherder.controller('PluginCtrl', [
...
-                        $scope.buildbotJobname = $scope.artifacts.buildapi.blob.buildername;
-                        setBuildernameHref($scope.buildbotJobname);
+                        $scope.jobLinkText = $scope.artifacts.buildapi.blob.buildername;
+                    } else {
+                        $scope.jobLinkText = $scope.job.job_type_name  + " " +
+                                             $scope.job.job_type_symbol ;
                     }
+                    $scope.searchStr = $scope.jobLinkText + " " +
+                                       $scope.job.signature;
+                    setJobSearchStrHref($scope.searchStr);
```
maurodoglio: Filtering on job type, symbol and signature seems a little bit redundant. What about using only the signature?

webapp/app/js/services/main.js
```diff
@@ -215,3 +215,20 @@ treeherder.factory('thNotify', [
     return thNotify;
 
 }]);
+
+treeherder.factory('thPlatformName', [
```
camd: This **was** only in ``clonejobs.js`` but needed to be shared elsewhere.  Namely ``plugins controller.js``.

webapp/app/plugins/pluginpanel.html
```diff
@@ -141,12 +141,21 @@
       <ul class="content-spacer">
         <li class="small">
           <label>Job:</label>
-          <a title="Filter jobs like these"
-             href="{{buildbotJobnameHref}}"
+          <a title="Filter by job data strings (or buildername)"
+             href="{{jobSearchStrHref}}"
              prevent-default-on-left-click
-             ng-click="filterByBuildername(buildbotJobname)"
```
camd: This isn't just for ``buildernames`` anymore.  Could also be fields from task cluster jobs.

webapp/app/plugins/pluginpanel.html
```diff
@@ -141,12 +141,21 @@
...
-             copy-value="{{buildbotJobname}}">
-            {{buildbotJobname}}</a>
+             ng-click="filterByJobSearchStr(jobSearchStr)"
+             copy-value="{{jobSearchStr}}">
+            {{jobSearchStr}}</a>
+        </li>
+        <li class="small">
```
camd: allow filtering by the specific job ``signature`` which is more accurate.  However, in discussions with sheriff's, they want both strings and this.

webapp/app/js/directives/clonejobs.js
```diff
@@ -539,17 +540,6 @@ treeherder.directive('thCloneJobs', [
         }
     };
 
-    var getPlatformName = function(name){
```
camd: moved this into a service so it could be shared

webapp/app/plugins/controller.js
```diff
@@ -10,20 +10,21 @@ treeherder.controller('PluginCtrl', [
...
+        thBuildApi, thNotify, ThJobLogUrlModel, ThModelErrors, thTabs,
+        $timeout, thPlatformName) {
 
         var $log = new ThLog("PluginCtrl");
 
         $scope.job = {};
         $scope.artifacts = {};
 
-        var setBuildernameHref = function(buildername){
+        var getJobSearchStrHref = function(jobSearchStr){
```
camd: made it a getter instead of a setter so it could be reused for the signature.

maurodoglio:
I meant to say that we could set $scope.jobSearchStr to the result of thJobSearchStr() no matter where the job comes from and forget about the buildername.
I'll leave the final call to you, I feel like I'm being very picky on the "buildername" topic

maurodoglio:
maybe @edmorley has an opinion on that?

edmorley:
So Mauro and I just chatted about this (we're both in the office) and I agree the best way to do this is:
* Stop using the buildername when clicking the link in the job details panel - and instead use a concatenated list of properties (or similar) to be able to pre-populate the filter box with something that will filter down to that job
* However retain (for now) the ability for someone to type a buildername manually in the filter field on the menu bar (or by specifying a &filter-searchStr=foo in the URL when say linking from another tool) that means for buildbot jobs they can still filter by buildername if they really need to

Does that sound reasonable? (I haven't looked at the PR diff yet)

tojonmz:
> Stop using the buildername when clicking the link in the job details panel

Assuming the hover copy-value on the Job Name would still remain intact, sounds good :)

camd:
@edmorley - Yep, sounds great.  My latest changes reflect this now.

PR #452, opened 2015-04-03
Bug 1151023 - Nojobs option

camd:
This fixes [Bug 1151023](https://bugzilla.mozilla.org/show_bug.cgi?id=1151023)

Add query string param to UI to skip loading jobs.  This makes it much faster when a user just wants to see the revisions in a large number of resultsets.

screenshot 2015-04-06 08 29 28

Help section update:

screenshot 2015-04-06 09 23 32

### DIFF
--- a/webapp/app/help.html
+++ b/webapp/app/help.html
@@ -546,6 +546,40 @@
         </div>
     </div>
 </div>
+<div class="col-xs-12">
+    <div class="panel panel-default">
+            <div class="panel-heading"><h3>URL Query String Parameters</h3></div>
+            <div class="panel-body panel-spacing">
+                <table id="queryparams">
+                    <tr>
+                      <td><span class="queryparam">nojobs</span></td>
+                      <td>Load result sets and revisions without loading any job results.</td>
+                      <td><span class="queryparam">&nojobs</span></td>
+                    </tr>
+                    <tr>
+                      <td><span class="queryparam">fromchange</span></td>
+                      <td>Specify the earliest revision hash in the resultset range.</td>
+                      <td><span class="queryparam">&fromchange=a12ca6c8b89b</span></td>
+                    </tr>
+                    <tr>
+                      <td><span class="queryparam">tochange</span></td>
+                      <td>Specify the latest revision hash in the resultset range.</td>
+                      <td><span class="queryparam">&tochange=3215c7fc090b</span></td>
+                    </tr>
+                    <tr>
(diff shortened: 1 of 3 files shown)


webapp/app/help.html
```diff
@@ -546,6 +546,35 @@
         </div>
     </div>
 </div>
+<div class="col-xs-12">
+    <div class="panel panel-default">
+            <div class="panel-heading"><h3>URL Query String Parameters</h3></div>
+            <div class="panel-body panel-spacing">
+                <table id="shortcuts">
```
tojonmz: I would maybe give this another id like 'queryparams'.

webapp/app/help.html
```diff
@@ -546,6 +546,35 @@
...
+                    <tr>
+                      <td><span class="kbd">nojobs</span></td>
```
tojonmz: I think we could omit the `kbd` class here since they aren't key shortcuts, and maybe use plain text for the 'nojobs' or '&nojobs' or whatever makes sense for the query param.

webapp/app/help.html
```diff
@@ -546,6 +546,35 @@
...
+                      <td><span class="kbd">fromchange</span></td>
+                      <td>Specify the earliest revision hash in the resultset range
+                    </tr>
+                    <tr>
+                      <td><span class="kbd">tochange</span></td>
+                      <td>Specify the latest revision hash in the resultset range
+                    </tr>
+                    <tr>
+                      <td><span class="kbd">startdate</span></td>
+                      <td>Specify the earliest date in the resultset range (2015-02-18)
```
tojonmz: Maybe we shouldn't hard code a date here, and below in the description.

But I agree the fromchange, tochange, startdate, endate could provide an example in the url param field on the left side of the table. For that date ones, maybe a "January 1st, 2015" to "January 1st, 2020" type of date range, so we don't have to keep updating the example.

tojonmz:
@camd integration testing with both FF and Chrome seems fine. I wonder if we should ng-disable the show/hide hidden jobs btn, or other applicable UI when the `&nojobs` param is present. Other things I noticed is the resultset "Complete" or job count doesn't get displayed with this param, which I guess is to be expected, but perhaps we could check and display a "No jobs loaded" or something in its place.

maurodoglio:
@camd the code looks good, I was just confused by a change in the second commit that seems to belong to the first one (webapp/app/js/models/resultsets_store.js
). I can live with that, but if you could move it to the first commit that would be great

PR #457, opened 2015-04-09
Bug 1149370 - Provide icons for repo btn states unsupported/not-retrieved

tojonmz:
This tweak fixes Bugzilla bug [1149370](https://bugzilla.mozilla.org/show_bug.cgi?id=1149370).

This provides an icon when the 'tree is unsupported', instead of just inheriting whatever is in the cache. While I was there I noticed we didn't have a state for 'not retrieved yet' either, so I vetted a couple of variants for that in channel, and am adding that also.

Here's a current repo in a happy state:

open

Here's 'tree is unsupported':

fa-question

Here's 'tree is not retrieved yet':

notretrievedyet2

For the purposes of the screen grabs, I only applied the new properties to the current open repo. To the extent possible without repos actually in that state, everything seems fine in local testing.

Tested on OSX 10.9.5:
FF Release **37.0.1**
Chrome Latest Release **41.0.2272.118** (64-bit)

Adding @edmorley for review.

### DIFF
--- a/webapp/app/js/directives/top_nav_bar.js
+++ b/webapp/app/js/directives/top_nav_bar.js
@@ -40,12 +40,12 @@ treeherder.directive('thWatchedRepo', [
             btnClass: "btn-view-nav-closed"
         },
         "unsupported": {
-            icon: "",
+            icon: "fa-question",
             color: "treeUnavailable",
             btnClass: "btn-view-nav"
         },
         "not retrieved yet": {
-            icon: "",
+            icon: "fa-spinner",
             color: "treeUnavailable",
             btnClass: "btn-view-nav"
         }


tojonmz:
And here was an alternate for 'tree is unsupported' in case we like exclamation better:

notretrievedyet1

Adding @KWierso for visibility.

tojonmz:
@KWierso has confirmed he prefers the ? for now, so we'll go with that for 'tree is unsupported'.

tojonmz:
Fixed by commit https://github.com/mozilla/treeherder-ui/commit/42b2f787f5342247bf1aaf677c6cede4b33ec027.

PR #459, opened 2015-04-09
Bug 1095745 - refresh failure summary when all logs parsed

camd:
Also stop updates if no job is de-selected (details panel closed)

### DIFF
--- a/webapp/app/plugins/controller.js
+++ b/webapp/app/plugins/controller.js
@@ -159,6 +159,10 @@ treeherder.controller('PluginCtrl', [
                                 }, 5000);
                             }
                         });
+                    } else {
+                        // if all logs are parsed, then update the failure summary tab
+                        // because there may now be failures or bug suggestions
+                        thTabs.tabs.failureSummary.update();
                     }
                     $scope.lvUrl = thUrl.getLogViewerUrl($scope.job.id);
                     $scope.lvFullUrl = location.origin + "/" + $scope.lvUrl;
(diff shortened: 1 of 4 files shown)


webapp/app/plugins/controller.js
```diff
@@ -138,7 +138,8 @@ treeherder.controller('PluginCtrl', [
                         }
                     });
                     // retry to fetch the job info if a log hasn't been parsed yet
-                    if(logsNotParsed.length > 0){
+                    // but only if a job is still selected
+                    if(logsNotParsed.length > 0 && $scope.selectedJob){
```
edmorley: I think the potential for a race (ie someone deselecting the job, so no job is selected) is more likely after this point, not before? ie: should the check for a selected job be wrapping selectJobAndRender() below - since that gets called after the 5 seconds? Or even better, could we cancel the setTimeout once the panel becomes deactivated?

webapp/app/plugins/controller.js
```diff
@@ -153,6 +154,10 @@ treeherder.controller('PluginCtrl', [
                                 }, 5000);
                             }
                         });
+                    } else {
+                        // if all logs are parsed, then update the failure summary tab
+                        // because there may now be failures or bug suggestions
+                        thTabs.tabs.failureSummary.update();
```
edmorley: Ah oops! Good spot :-)

camd:
@edmorley - yep, you're right.  Canceling the promises is a much better way to go.  Now it's immediate.

PR #461, opened 2015-04-13
Bug 1153834 - Move job description tooltips to job details panel

edmorley:
Instead of displaying the long job description in the tooltip for every job listed in the main jobs list, instead add it as the title text for the job name in the job details panel. This avoids bloating the DOM with thousands of longer strings.

We no longer check for the placeholder string "fill me", but I don't think this matters, since we shouldn't be using that string in my opinion anyway (vs leaving it blank) and to me it's more consistent to always show a tooltip, rather than have the user wonder where it has gone.

### DIFF
--- a/webapp/app/js/directives/clonejobs.js
+++ b/webapp/app/js/directives/clonejobs.js
@@ -56,19 +56,11 @@ treeherder.directive('thCloneJobs', [
         };
 
     var getHoverText = function(job) {
-
-        var jobStatus = thResultStatus(job);
-        var hoverText = job.job_type_name + " - " + jobStatus;
-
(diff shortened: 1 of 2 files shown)


tojonmz:
I just tested the branch for interest, it looks good so far. I take it sheriffs are on board with the move.

I agree, perhaps subsequently or with this change, we provide a string that says "No description provided" or something in the tooltip, rather than 'fill me'.

fillmn-e10s

edmorley:
> I agree, perhaps subsequently or with this change, we provide a string that says "No description provided" or something in the tooltip, rather than 'fill me'.

Yeah let's deal with that in bug 1052501 :-)

PR #465, opened 2015-04-14
initial work to get compare-talos stood up inside of perfherder

jmaher:
first round attempt.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/465)


edmorley:
Thank you for taking a look at this :-)
The PR will also need changes to https://github.com/mozilla/treeherder-ui/blob/master/Gruntfile.js (use eg the perf.html entries as examples).

webapp/app/compare.html
```diff
@@ -0,0 +1,39 @@
...
+    </style>
+</head>
+<body>
+
+  <section ui-view>
+  </section>
+
+  <script src="js/config/local.conf.js"></script>
+
+  <!-- build:js js/perf.min.js -->
```
wlach: This will override the perf.js file. Let's change perf.min.js -> compare.min.js. As @edmorley pointed out, you'll also need to make some gruntfile changes for the minification to work in production.

webapp/app/js/compare.js
```diff
@@ -0,0 +1,494 @@
...
+          .reduce(function (a, b) { return a + b; }) / (values.length - 1));
+}
+
+
+// -------------------------------------------------------------------------
+// End Utility Functions
+// -------------------------------------------------------------------------
+
+
+var comparetalos = angular.module("perf", ['ui.router', 'ui.bootstrap']);
```
wlach: comparetalos -> compareTalos

wlach: Also, add 'treeherder' to the dependencies list, so: ['ui.router', 'ui.bootstrap', 'treeherder']

webapp/app/js/compare.js
```diff
@@ -0,0 +1,494 @@
...
+/* Copied from providers.js */
+comparetalos.provider('thServiceDomain', function() {
+    this.$get = function() {
+        if (window.thServiceDomain) {
+            return window.thServiceDomain;
+        } else {
+            return "";
+        }
+    };
+});
```
wlach: This isn't needed now that bug 1152376 has landed. You can just pull this code out (it will be pulled in as part of the treeherder module).

webapp/app/js/compare.js
```diff
@@ -0,0 +1,494 @@
...
+
+comparetalos.factory('getSeriesSummary', [ function() {
```
wlach: We should create a higher-level perfherder module to include functions like this (and rename the existing perf module to perfherder.graphs). This is almost identical to the same function in perf.js.

webapp/app/js/compare.js
```diff
@@ -0,0 +1,494 @@
...
+    var availableColors = [ 'red', 'green', 'blue', 'orange', 'purple' ];
+
+    function displayComparision() {
+//TODO: why do we need so much history?
+      $scope.timeRange = 2592000; // last 30 days
+      $scope.testList = [];
+
+//TODO: how to input these variables
+$scope.e10s = false;
+$scope.pgo = false;
```
wlach: You can get them via the $stateParams variable, see line 473 in this file. https://github.com/mozilla/treeherder-ui/pull/465/files#diff-921250c6e3d3318db57e5548862c72faR473

P.S. You only want to define variables as members of $scope if you want to bind them back to the view, which AFAIK in this case you do not.

webapp/app/js/compare.js
```diff
@@ -0,0 +1,494 @@
...
+    if (testName === undefined) {
+      testName = "summary";
+      subtestSignatures = signatureProps.subtest_signatures;
+    }
+    var name = signatureProps.suite + " " + testName +
+      " " + optionCollectionMap[signatureProps.option_collection_hash] + extra;
+
+    //Only keep summary signatures, filter in/out e10s and pgo
+    if (name.indexOf('summary') <= 0) {
+        return {};
```
wlach: I would return null for these cases rather than a blank object ({}).

webapp/app/js/compare.js
```diff
@@ -0,0 +1,494 @@
...
+
+          // TODO: figure how how to reduce these maps
+          var rawResultsMap = {};
+
+          $http.get(uri2).then(function(response) {
+            response.data.forEach(function(data) {
+              var perfData = data.blob;
+              var geomeans = [];
+              var total = 0;
+              for (var i=0; i < perfData.length; i++) {
```
wlach: use forEach here, not array indexes.

webapp/app/js/compare.js
```diff
@@ -0,0 +1,494 @@
...
+              return;
+            }
+
+            var testname = seriesSummary.name;
+            if ($scope.testList.indexOf(testname) === -1) {
+              $scope.testList.push(testname);
+            }
+
+            // Bug 1153301, query too long
+            if (seriesSummary.signature !== undefined && selectedSignaturesCount < 20) {
```
wlach: As I mentioned earlier, we can batch requests to get this information and wait on all of them be complete via $q promises. See this example in perf.js: https://github.com/mozilla/treeherder-ui/blob/master/webapp/app/js/perf.js#L524

webapp/app/js/compare.js
```diff
@@ -0,0 +1,494 @@
...
+                    }
+
+                    geomeans.push(perfData[i].geomean);
+                    total += perfData[i].geomean;
+                  }
+
+                  var avg = total / geomeans.length;
+                  var sigma = stddev(geomeans, avg);
+                  new_rawResultsMap[data.series_signature] = {'geomean': avg, 'variation': 2*sigma, 'runs': geomeans.length};
+                });
```
wlach: You really need to factor this duplicated code to get results for prev/next revisions into a common function and call it twice. :)

webapp/app/js/compare.js
```diff
@@ -0,0 +1,494 @@
...
+
+
+
+
+              var counter = 0;
+              for (var t in $scope.testList) {
```
wlach: Use forEach here

webapp/app/js/compare.js
```diff
@@ -0,0 +1,494 @@
...
+                // Remove previous header if we have no data
+                if (counter > 0 && compareResultsMap[(counter-1)].isHeader) {
+                  counter--;
+                }
+
+                //TODO: figure out a cleaner method for making the names a header row
+                compareResultsMap[counter++] = {'name': $scope.testList[t], 'isHeader': true};
+                //TODO; shouldn't this be on new_seriesList and is idx from above == idx here?
+                for (var idx in seriesList) {
```
wlach: again, use forEach

webapp/app/js/compare.js
```diff
@@ -0,0 +1,494 @@
...
+          '/resultset/?format=json&full=false&with_jobs=false&revision=' + 
+          $scope.newRevision;
+
+      $http.get(uri).then(function(response) {
+        var results = response.data.results;
+        if (results.length > 0) {
+          $scope.newResultSetID = results[0].id;
+        }
+      });
+    }
```
wlach: verifyOriginal and verifyNew are almost the same function. You should refactor them into one and call it twice.

webapp/app/partials/perf/comparectrl.html
```diff
@@ -0,0 +1,44 @@
...
+              <th></th><th>Test/Platform</th><th>Original</th><th>Variation</th><th>New</th><th>Variation</th><th>Delta</th><th>Delta %</th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr ng-class="{'subtestheader': compareResult.isHeader}" ng-repeat="compareResult in compareResults">
+              <td><a ng-style="compareResult.isHeader && {'display': 'none'}" ng-href="{{compareResult.detailsLink}}">Details</a></td>
+              <td>{{compareResult.name}}</td>
+              <td ng-attr-title="runs: {{compareResult.originalRuns}}">{{compareResult.originalGeoMean}}</td><td>{{compareResult.originalVariation}}</td>
+              <td ng-attr-title="runs: {{compareResult.newRuns}}">{{compareResult.newGeoMean}}</td><td>{{compareResult.newVariation}}</td>
+              <td ng-style="compareResult.type === 'regression' && {'background-color': 'red', 'color': 'white', 'font-weight': 'bold'} ||
```
wlach: We should define these as classes in perf.css, instead of setting raw CSS values conditionally in the html. That makes it easier for people who are of more of a web designer bent to contribute.

webapp/app/compare.html
```diff
@@ -0,0 +1,39 @@
+<!DOCTYPE html>
+<html ng-app="perf">
+<head>
+    <meta charset="iso-8859-1" />
+    <title>Perfherder Graphs</title>
```
wlach: This probably needs some kind of new title. :)

avih:
If there was an online version where the progress could be tested (rather than only after you declare it done), it could help others give feedback.

jmaher:
avih you can clone my repo and run it locally- it is very easy!  I don't think wlach is going for perfection here, just basics to get it in and we can iterate on it.

edmorley:
@avih, install nodejs then just:
https://treeherder-ui.readthedocs.org/en/latest/installation.html#running-the-web-server

avih:
Maybe it will and maybe it won't work out of the box (and it'd also need figuring out how to operate it, where to point various bits, etc), but a link to a live preview of a specific changeset with the thing which you want feedback for right at the front - would help a lot for people who don't work on this regularly.

avih:
Is there a staging server where these changesets could be deployed and examined in a relatively quick cycle?

jmaher:
Avi.  This is a pull request.  It needs some type of sanity review before
getting to the staging server.  When this is landed it will be about half
way done.

Joel

edmorley:
@jmaher, you could also clone the branch onto a people account (and cp the example config file, which points at prod's API) if you wanted feedback before merging - it should Just Work from there too.

webapp/app/compare.html
```diff
@@ -0,0 +1,40 @@
+<!DOCTYPE html>
```
wlach: Let's call this file compareperf.html

webapp/app/css/perf.css
```diff
@@ -165,6 +165,11 @@ to { opacity: 0; }
 }
 
 
+.subtestheader {
```
wlach: Let's call this .subtest-header

webapp/app/js/compare.js
```diff
@@ -0,0 +1,448 @@
...
+
+    return { name: name, signature: signature, platform: platform,
+             subtestSignatures: subtestSignatures };
+  };
+}]);
+
+compare.controller('CompareCtrl', [ '$state', '$stateParams', '$scope', '$rootScope', '$location',
+                              '$modal', 'thServiceDomain', '$http', '$q', '$timeout', 'getSeriesSummary',
+  function CompareCtrl($state, $stateParams, $scope, $rootScope, $location, $modal,
+                    thServiceDomain, $http, $q, $timeout, getSeriesSummary) {
```
wlach: As discussed on irc, let's leave out the modal stuff for now. You can just remove $modal from the list of dependencies here to inject.

webapp/app/js/compare.js
```diff
@@ -0,0 +1,448 @@
...
+                           'get_performance_data/?interval_seconds=' + $scope.timeRange;
+
+            //ok, now get the new revision
+            var signatureListURL = thServiceDomain + '/api/project/' + $scope.newProject + 
+              '/performance-data/0/get_performance_series_summary/?interval=' +
+              $scope.timeRange;
+
+            // TODO: figure how how to reduce these maps
+            var new_rawResultsMap = {};
+            var new_seriesList = [];
```
wlach: No underscores in variable names, use newRawResultsMap, etc.

webapp/app/js/compare.js
```diff
@@ -0,0 +1,448 @@
...
+          });
+        }
+      );
+    }
+
+    function calculateStats(perfData, resultSetID) {
+      var geomeans = [];
+      var total = 0;
+      perfData.forEach(function(pdata) {
+        if (pdata.result_set_id != resultSetID) {
```
wlach: Change this to if (pdata.result_set_id === resultSetID) { , then push the geomean/total updating stuff in here. That way we can avoid returning null.

webapp/app/js/compare.js
```diff
@@ -0,0 +1,448 @@
...
+          }
+
+          var cmap = {};
+          cmap['originalGeoMean'] = rawResultsMap;
+          cmap['newGeoMean'] = new_rawResultsMap;
+
+          if (seriesList[idx].signature in rawResultsMap) {
+             cmap['originalGeoMean'] = rawResultsMap[seriesList[idx].signature].geomean.toFixed(2);
+             cmap['originalRuns'] = rawResultsMap[seriesList[idx].signature].runs;
+//TODO: \b1 doesn't print out the +- character
```
wlach: Should indent in line with the code.

webapp/app/js/compare.js
```diff
@@ -0,0 +1,448 @@
...
+
+        };
+      });
+      $scope.compareResults = Object.keys(compareResultsMap).map(function(k) {
+        return compareResultsMap[k];
+      });
+    }
+
+    function verifyRevision(project, revision, rsid) {
+      if ((revision != null && revision != '') &&
```
wlach: This if clause looks wrong, we should have already verified we have data to process when we reach this point.

webapp/app/js/compare.js
```diff
@@ -0,0 +1,448 @@
...
+          $stateParams.pgo = true;
+        } else {
+          $stateParams.pgo = false;
+        }
+
+        // TODO: validate projects and revisions
+        $scope.originalProject = '';
+        if ($stateParams.originalProject) {
+          $scope.originalProject = $stateParams.originalProject;
+        }
```
wlach: I don't understand the point of all this if stuff. Why not just say $scope.originalProject = $stateParams.originalProject; ?

webapp/app/js/compare.js
```diff
@@ -0,0 +1,448 @@
...
+        $scope.newRevision = '';
+        if ($stateParams.newRevision) {
+          $scope.newRevision = $stateParams.newRevision;
+        }
+
+        $scope.originalRevision = '';
+        if ($stateParams.originalRevision) {
+          $scope.originalRevision = $stateParams.originalRevision;
+        }
+
```
wlach: For now, let's do a check here to make sure we have all the variables we need to proceed here. If we don't, just return (maybe set a $scope.insufficientData variable so we can display an error message?)

webapp/app/js/compare.js
```diff
@@ -0,0 +1,448 @@
...
+        verifyRevision($scope.originalProject, $scope.originalRevision, "original");
+        verifyRevision($scope.newProject, $scope.newRevision, "new");
+
+        $http.get(thServiceDomain + '/api/repository/').then(function(response) {
+          $scope.projects = response.data;
+
+          $scope.addCompareData = function() {
```
wlach: Let's remove this for now.

webapp/app/js/compare.js
```diff
@@ -0,0 +1,448 @@
...
+  };
+}]);
+
+compare.controller('CompareCtrl', [ '$state', '$stateParams', '$scope', '$rootScope', '$location',
+                              '$modal', 'thServiceDomain', '$http', '$q', '$timeout', 'getSeriesSummary',
+  function CompareCtrl($state, $stateParams, $scope, $rootScope, $location, $modal,
+                    thServiceDomain, $http, $q, $timeout, getSeriesSummary) {
+
+    function displayComparision() {
+      //TODO: why do we need so much history?
```
wlach: I assume we need the history so we can dig out older results? We could probably do better by seeing how old each revision is and only fetching enough data to cover them both.

webapp/app/js/compare.js
```diff
@@ -0,0 +1,448 @@
...
+        if (counter > 0 && compareResultsMap[(counter-1)].isHeader) {
+          counter--;
+        }
+
+        //TODO: figure out a cleaner method for making the names a header row
+        compareResultsMap[counter++] = {'name': testName, 'isHeader': true};
+
+        //TODO: need to figure out how to iterate through the old and new series list
+        //      Ideally a seriesList.forEach(...)
+        for (var idx in seriesList) {
```
wlach: Can we not use forEach here as well?

webapp/app/js/compare.js
```diff
@@ -0,0 +1,448 @@
...
+          if (seriesList[idx].name != testName) {
+            continue;
+          }
+
+          var cmap = {};
+          cmap['originalGeoMean'] = rawResultsMap;
+          cmap['newGeoMean'] = new_rawResultsMap;
```
wlach: You could use a cmap.newGeoMean syntax here. A bit more concise.

webapp/app/js/compare.js
wlach: Actually wait, never mind, I see why you did that. Maybe there's something in underscore.js that will fit the bill though...

wlach: Yup, you could use underscore's zip function and still use forEach here: http://underscorejs.org/#zip

webapp/app/js/compare.js
```diff
@@ -0,0 +1,448 @@
...
+
+    function verifyRevision(project, revision, rsid) {
+      if ((revision != null && revision != '') &&
+          (project != null && project != '')) {
+
+        var uri = thServiceDomain + '/api/project/' + project +
+            '/resultset/?format=json&full=false&with_jobs=false&revision=' + 
+            revision;
+
+        $http.get(uri).then(function(response) {
```
wlach: To get a promise here, you can just put a return statement in front of $http.get

webapp/app/js/compare.js
```diff
@@ -0,0 +1,448 @@
...
+          $stateParams.e10s = true;
+        } else {
+          $stateParams.e10s = false;
+        }
+
+        if ($stateParams.pgo) {
+          $stateParams.pgo = true;
+        } else {
+          $stateParams.pgo = false;
+        }
```
wlach: As far as I can tell these if clauses don't do anything? If you really want a boolean value, do something like $stateParams.pgo = Boolean($stateParams.pgo), which is at least less lines of code. I don't see the need though.

webapp/app/js/compare.js
```diff
@@ -0,0 +1,448 @@
...
+    $scope.newProject = _.findWhere(projects, {name: defaultProjectName});
+  } else {
+    $scope.newProject = 'Try';
+  }
+
+  $scope.loadingCompareData = false;
+  $scope.originalRevision = '';
+  $scope.newRevision = '';
+
+  $scope.addCompareData = function () {
```
wlach: We can remove this.

webapp/app/js/compare.js
```diff
@@ -0,0 +1,448 @@
...
+    $modalInstance.close($scope.orignalProject, $scope.originalRevision,
+                         $scope.newProject, $scope.newRevision);
+  };
+
+  $scope.cancel = function () {
```
wlach: Also this.

webapp/app/partials/perf/comparedatachooser.html
```diff
@@ -0,0 +1,21 @@
+<div class="modal-header">
```
wlach: As discussed, let's leave this out for now.

webapp/app/partials/perf/comparectrl.html
wlach: You want to use ng-class for this, not ng-style https://docs.angularjs.org/api/ng/directive/ngClass

webapp/app/compareperf.html
```diff
@@ -0,0 +1,40 @@
+<!DOCTYPE html>
+<html ng-app="compare">
+<head>
+    <meta charset="iso-8859-1" />
+    <title>Compare Talos</title>
```
wlach: Let's call this `Compare Performance` as we'll be using this for things aside from Talos at some point likely in the not too distant future.

webapp/app/css/perf.css
```diff
@@ -165,6 +165,25 @@ to { opacity: 0; }
 }
 
 
+.subtest-header {
+    background-image: linear-gradient(to bottom, #6A7B86, #424F5A);
+    color: white;
+    font-weight: bold;
+}
```
wlach: Blank lines between CSS declarations plz.

webapp/app/partials/perf/comparectrl.html
```diff
@@ -0,0 +1,42 @@
+<nav class="navbar navbar-default" role="navigation">
+  <div class="container-fluid">
+    <div class="navbar-header">
+      <a class="navbar-brand" href="#">Perfherder Compare Talos</a>
+    </div>
+  </div>
+</nav>
+
+<div class="ph-horizontal-layout">
+  <div id="revision-chooser">
```
wlach: Remove this div

webapp/app/partials/perf/comparectrl.html
```diff
@@ -0,0 +1,42 @@
+<nav class="navbar navbar-default" role="navigation">
+  <div class="container-fluid">
+    <div class="navbar-header">
+      <a class="navbar-brand" href="#">Perfherder Compare Talos</a>
+    </div>
+  </div>
+</nav>
+
+<div class="ph-horizontal-layout">
```
wlach: Remove the ph-horizontal-layout class, replace with "container-fluid". We don't need flexbox behaviour anymore, and using container-fluid will give us some nice padding.

webapp/app/partials/perf/comparectrl.html
```diff
@@ -0,0 +1,42 @@
...
+  <div id="revision-chooser">
+   &nbsp;
+  </div>
+  <div id="data-display">
```
wlach: You can remove this container div now.

webapp/app/partials/perf/comparectrl.html
```diff
@@ -0,0 +1,42 @@
...
+              <td>+/-{{compareResult.newStddev}} ({{compareResult.newStddevPct}}%)</td>
+              <td ng-class="{'subtest-regression': compareResult.isRegression, 'subtest-improvement':  compareResult.isImprovement}">{{compareResult.delta}}</td>
+              <td ng-class="{'subtest-regression': compareResult.isRegression, 'subtest-improvement': compareResult.isImprovement}">{{compareResult.deltaPercentage}}%</td>
+            </tr>
+          </tbody>
+        </table>
+     </div>
+    </div>
+  </div>
+  <div class="graph-right-padding"></div>
```
wlach: You can remove this div

webapp/app/partials/perf/comparectrl.html
```diff
@@ -0,0 +1,42 @@
+<nav class="navbar navbar-default" role="navigation">
+  <div class="container-fluid">
+    <div class="navbar-header">
+      <a class="navbar-brand" href="#">Perfherder Compare Talos</a>
```
wlach: Likewise, let's rename this to "Perfherder Compare"

webapp/app/partials/perf/comparectrl.html
```diff
@@ -0,0 +1,42 @@
...
+   &nbsp;
+  </div>
+  <div id="data-display">
+    <div id="datapoint-detail" ng-show="originalRevision && newRevision">
+      <div ng-if="dataLoading">
+        Loading all results, please wait a minute...
+        <img src="img/dancing_cat.gif" />
+      </div>
+      <div id="subtest-summary" ng-if="!dataLoading">
+        <h4>Compare All Data for Revision {{newProject}} : {{newRevision}} to {{originalProject}} : {{originalRevision}}</h4>
```
wlach: Some spacing and capitalization issues here. I'd prefer:

Compare all data for revision {{newProject}}: to {{originalProject}}: {{originalRevision}}

webapp/app/js/compare.js
```diff
@@ -0,0 +1,403 @@
...
+        $scope.originalProject = $stateParams.originalProject;
+        $scope.newProject = $stateParams.newProject;
+        $scope.newRevision = $stateParams.newRevision;
+        $scope.originalRevision = $stateParams.originalRevision;
+        if ($scope.originalProject == '' ||
+            $scope.newProject == '' ||
+            $scope.originalRevision == '' ||
+            $scope.newRevision == '') {
+          //TODO: get an error to the UI
+        }
```
wlach: You probably want to return here.

webapp/app/js/compare.js
```diff
@@ -0,0 +1,403 @@
...
+        $stateParams.pgo = Boolean($stateParams.pgo);
+        $stateParams.e10s = Boolean($stateParams.e10s);
+        $scope.hideMinorChanges = Boolean($stateParams.hideMinorChanges);
+
+        // TODO: validate projects and revisions
+        $scope.originalProject = $stateParams.originalProject;
+        $scope.newProject = $stateParams.newProject;
+        $scope.newRevision = $stateParams.newRevision;
+        $scope.originalRevision = $stateParams.originalRevision;
+        if ($scope.originalProject == '' ||
```
wlach: Replace $scope.foo == '' with !$scope.foo

webapp/app/js/compare.js
```diff
@@ -0,0 +1,403 @@
...
+                           'get_performance_data/?interval_seconds=' + $scope.timeRange;
+
+            //ok, now get the new revision
+            var signatureListURL = thServiceDomain + '/api/project/' + $scope.newProject + 
+              '/performance-data/0/get_performance_series_summary/?interval=' +
+              $scope.timeRange;
+
+            var newRawResultsMap = {};
+            var newSeriesList = [];
+
```
wlach: Remove one newline here.

webapp/app/js/compare.js
```diff
@@ -0,0 +1,403 @@
...
+          $q.all(seriesList.map(function(series) {
+            return $http.get(signatureURL + "&signatures=" + series.signature).then(function(response) {
+              response.data.forEach(function(data) {
+                rawResultsMap[data.series_signature] = calculateStats(data.blob, $scope.originalResultSetID);
+                rawResultsMap[data.series_signature].name = series.name;
+                rawResultsMap[data.series_signature].platform = series.platform;
+              });
+            });
+          })).then(function () {
+
```
wlach: Superfluous newline

webapp/app/js/compare.js
```diff
@@ -0,0 +1,403 @@
...
+                  }
+
+                  if ($scope.testList.indexOf(seriesSummary.name) === -1) {
+                    $scope.testList.push(seriesSummary.name);
+                  }
+                }
+              });
+              $scope.testList.sort();
+              $scope.platformList.sort();
+
```
wlach: Remove one newline here.

webapp/app/js/compare.js
```diff
@@ -0,0 +1,403 @@
...
+          });
+        }
+      );
+    }
+
+    //TODO: put this into a generic library
+    function calculateStats(perfData, resultSetID) {
+      var geomeans = [];
+      var total = 0;
+      perfData.forEach(function(pdata) {
```
wlach: Actually now that I think about it we could express this even more concisely with underscore.js:

_.where(perfData, { result_set_id: resultSetId }).forEach(function(pdata) {
...
}

http://underscorejs.org/#where

webapp/app/js/compare.js
```diff
@@ -0,0 +1,403 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
wlach: Let's call this compareperf.js

webapp/app/js/compare.js
```diff
@@ -0,0 +1,403 @@
...
+      perfData.forEach(function(pdata) {
+        if (pdata.result_set_id === resultSetID) {
+          geomeans.push(pdata.geomean);
+          total += pdata.geomean;
+        }
+      });
+
+      var avg = total / geomeans.length;
+      var sigma = stddev(geomeans, avg);
+      return {'geomean': avg.toFixed(2), 'stddev': sigma.toFixed(2), 'runs': geomeans.length};
```
wlach: you don't need the quotes around the object members here

webapp/app/js/compare.js
```diff
@@ -0,0 +1,403 @@
...
+    function displayResults(rawResultsMap, newRawResultsMap) {
+      var counter = 0;
+      var compareResultsMap = {};
+
+      $scope.testList.forEach(function(testName) {
+        if (counter > 0 && compareResultsMap[(counter-1)].headerColumns==2) {
+          counter--;
+        }
+
+        //TODO: figure out a cleaner method for making the names a header row
```
wlach: I think it would honestly be better to just hardcode most of the header inside the .html. For the stuff that changes, let's store a bunch of seperate properties and modify the header inside the html accordingly. Then just do the ng-repeat on the individual rows.

I guess this would mean also breaking out each of the subgroups, but that shouldn't be so bad, I don't think? I'm imagining an object like this:

$scope.compareGroups = [ { header: { ... header properties ... }, { data: { ... data properties ... }, ... ]

webapp/app/js/compare.js
```diff
@@ -0,0 +1,403 @@
...
+                      'newGeoMean': NaN, 'newRuns': 0, 'newStddev': NaN, 'headerColumns': 1,
+                      'delta': NaN, 'deltaPercentage': NaN, 'isEmpty': false,
+                      'isRegression': false, 'isImprovement': false, 'isMinor': true};
+
+          var oldSig = _.find(Object.keys(rawResultsMap), function (sig) {
+            return (rawResultsMap[sig].name == testName && rawResultsMap[sig].platform == platform)});
+          var newSig = _.find(Object.keys(newRawResultsMap), function (sig) {
+            return (newRawResultsMap[sig].name == testName && newRawResultsMap[sig].platform == platform)});
+
+          if (oldSig !== undefined) {
```
wlach: !== undefined isn't necessary, I don't think

wlach:
Merged manually.

PR #468, opened 2015-04-14
Bug 1151337-Perfherder graphs should use OptionCollection model

git-srinivas:
Replaced http.get call with option collection model service call in perf module

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/468)

### DIFF
--- a/webapp/app/js/perf.js
+++ b/webapp/app/js/perf.js
@@ -30,9 +30,9 @@ perf.factory('getSeriesSummary', [ function() {
 }]);
 
 perf.controller('PerfCtrl', [ '$state', '$stateParams', '$scope', '$rootScope', '$location',
-                              '$modal', 'thServiceDomain', '$http', '$q', '$timeout', 'getSeriesSummary',
+                              '$modal', 'thServiceDomain', '$http', '$q', '$timeout', 'getSeriesSummary', 'ThOptionCollectionModel',
   function PerfCtrl($state, $stateParams, $scope, $rootScope, $location, $modal,
-                    thServiceDomain, $http, $q, $timeout, getSeriesSummary) {
+                    thServiceDomain, $http, $q, $timeout, getSeriesSummary, ThOptionCollectionModel) {
 
     var availableColors = [ 'red', 'green', 'blue', 'orange', 'purple' ];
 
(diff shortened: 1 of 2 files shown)


webapp/app/perf.html
```diff
@@ -36,6 +36,9 @@
   <script src="vendor/mousetrap.min.js"></script>
   <script src="js/treeherder.js"></script>
   <script src="js/providers.js"></script>
+  <script src="js/models/option_collection.js"></script>
+  <script src="js/services/main.js"></script>
+  <script src="js/services/log.js"></script>
```
wlach: I don't understand why you're including the last two lines.

webapp/app/js/perf.js
```diff
@@ -650,13 +650,16 @@ perf.controller('PerfCtrl', [ '$state', '$stateParams', '$scope', '$rootScope',
...
-        response.data.forEach(function(dict) {
-          optionCollectionMap[dict.option_collection_hash] =
-            dict.options.map(function(option) {
-              return option.name; }).join(" ");
-        });
+    ThOptionCollectionModel.get_list()
+      .success(function(optCollectionData) {
+                 // gather the string representations of option collections
+                 
+                 _.each(optCollectionData, function(optColl) {
```
wlach: The indentation on these lines isn't correct -- it's too deep and inconsistent with the rest of the file. Please follow the style of the original function.

wlach:
This looks ok except for a few stray imports and indentation problems. Does it seem to work ok in your testing? If so, please fix the few issues I mention, amend the commit and push to the fork-- then we can get this merged.

Also, in the future please file review requests for these as outlined in http://ateam-bootcamp.readthedocs.org/en/latest/guide/development_process.html#git-and-github

If you follow the process there it is likely your patch will be looked into sooner.

webapp/app/perf.html
git-srinivas: OptionCollection model uses  thUrl service  which is in main.js  and thUrl service is using  ThLog, it is in log.js. They are the services being used by option collection model.

webapp/app/js/perf.js
git-srinivas: Ok lach, i will correct the indentation :)

PR #470, opened 2015-04-15
Bug 1148664 - Limit notifications from blocking jobs and other UI - do not merge

tojonmz:
This tweak addresses the main user problem in Bugzilla bug [1148664](https://bugzilla.mozilla.org/show_bug.cgi?id=1148664).

Perhaps we could economize on dimension, location, and frequency of the notifications later, but with this change at least the notifications on retrigger don't obscure the pinboard and job panel. So in z-index I set:

* notifications at 500
* the entire job panel at 1000 (not just its navbar)
* pinboard overlay menus at 1100 (was 2000)

The upper Filter menu still lies *below* notifications, and the top navbar menus still draw above them. No changes there.

Here's the before:

current

Here's after:

proposed

Everything seems fine with various UI, but it will be worth everyone trying it on dev/stage in case there's any adjustments to be made.

Tested on OSX 10.9.5:
FF Release **37.0.1**
Chrome Latest Release **41.0.2272.118** (64-bit)

Adding @wlach for review.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/470)

### DIFF
--- a/webapp/app/css/treeherder.css
+++ b/webapp/app/css/treeherder.css
@@ -610,6 +610,7 @@ div#bottom-panel {
     display:         flex;
     -webkit-flex-flow: column;
     flex-flow: column;
+    z-index: 1000;
 }
 
 div#bottom-panel .navbar{
@@ -623,7 +624,6 @@ div#bottom-panel .navbar{
     min-height: 33px;
     min-width: initial;
     overflow: hidden;
-    z-index: 100;
 }
 
 div#bottom-panel .navbar-nav > ul{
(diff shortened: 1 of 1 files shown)


tojonmz:
We are going to expand the scope of the fix with a max-notification limit. This should be particularly helpful for retriggers.

tojonmz:
So the nature of the proposed fix is so different (and my local branch name reflects that difference) I'm going to close this PR and open a new one.

PR #473, opened 2015-04-17
Bug 1155321 - fix filter params for jobs not in a group

camd:
If a job (like a build) is not in a group, then the job details string
for filtering by fields showed “unknown ?” as the job-group name and
symbol.  Those are legitimate values internally, but the user shouldn’t
see them.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/473)

### DIFF
--- a/webapp/app/js/services/main.js
+++ b/webapp/app/js/services/main.js
@@ -231,8 +231,10 @@ treeherder.factory('thJobSearchStr', [
 
         return function(job) {
             return thPlatformName(job.platform) + ' ' +
-                job.platform_option + ' ' + job.job_group_name + ' ' +
-                job.job_group_symbol + ' ' + job.job_type_name + ' ' +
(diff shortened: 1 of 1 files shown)


webapp/app/js/services/main.js
```diff
@@ -231,8 +231,10 @@ treeherder.factory('thJobSearchStr', [
 
         return function(job) {
             return thPlatformName(job.platform) + ' ' +
-                job.platform_option + ' ' + job.job_group_name + ' ' +
-                job.job_group_symbol + ' ' + job.job_type_name + ' ' +
+                job.platform_option + ' ' +
+                (job.job_group_name === 'unknown'? '': job.job_group_name + ' ') +
```
edmorley: nit: Spaces around the operators please :-)

PR #483, opened 2015-04-22
Bug 1148664 - Expire non-sticky notifications with a max queue

tojonmz:
This work fixes Bugzilla bug [1148664](https://bugzilla.mozilla.org/show_bug.cgi?id=1148664).

This prevents the UX problem of sequential notifications from tiling down the page and blocking the jobs and job panel. Specifically we now:

* enforce a maximum number of notifications (proposing 5)
* relocate them to the left side over the revision table so we don't obscure jobs
* expire/prune 'old' non-sticky notifications when new ones arrive
* reduce the height of all notifications
* align the 'X" close icon for sticky notifications to the text baseline
* use a new yellow 'warning' notification if the user retriggers before the job is loaded
* inject the job type (eg. 'B', 'Cpp') into retrigger success messages since the notifications cascade
* make the thModelErrors catch notification non-sticky
* support sticky notifications (will appear at the 'top') with co-incident non-sticky ('below', cascading)

There's lots of behaviors to check and test, some of them:

* retrigger without being logged in
* retrigger without LDAP access (my situation here)
* retrigger with full access (LDAP access)
* issue rapid retriggers before the job loads (should produce our new yellow warning message)
* issue other notifications (eg. Save Classification, PinAll, etc) in combination
* issue persistent/sticky and then issue non-sticky notifications

I've done a lot of testing, and everything seems fine so far. I wanted to use `$timeout(thNotify.shift)` directly in the for loop which prunes the non-sticky notifications, but during stress testing with many rapid notifications, it appears that `$timeout(thNotify.shift)` will start to do weird things, pruning the queue where the 4000ms timeout should not have occurred. So I chose to duplicate the `shift` code, which appears to work perfectly.

So here's what we used to have (imagine for the sake of my screen grab these were successes):

current

Here's is proposed (simulating a max of 5 retriggers I didn't want to actually do):

proposed

Here is an example of a successful retrigger with job type identifier:

successfulretrigger

Here is an edge case showing sequential retriggers before the job has loaded (it's typically sporadic and mixed in with successes):

notyetloadedproposed

There's all sorts of fun variations you can try in testing, but everything seems to be behaving as I expect, on both browsers. Sheriffs have signed off on all behaviors and re-location.

Tested on OSX 10.9.5:
FF Release **37.0.2**
Chrome Latest Release **42.0.2311.90** (64-bit)

Adding @camd for review and @maurodoglio for visibility.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/483)

### DIFF
--- a/webapp/app/js/services/main.js
+++ b/webapp/app/js/services/main.js
@@ -184,12 +184,18 @@ treeherder.factory('thNotify', [
             $log.debug("received message", message);
             var severity = severity || 'info';
             var sticky = sticky || false;
+            var maxNsNotifications = 5;
             thNotify.notifications.push({
                 message: message,
                 severity: severity,
                 sticky: sticky
             });
-            if(!sticky){
+
+            if (!sticky) {
+                if (thNotify.notifications.length > maxNsNotifications) {
+                    $timeout(thNotify.shift);
+                    return;
+                }
                 $timeout(thNotify.shift, 4000, true);
             }
         },
--- a/webapp/app/css/treeherder.css
+++ b/webapp/app/css/treeherder.css
@@ -1912,18 +1912,24 @@ div.similar_jobs .left_panel table tr.active > td {
 /* Avoid using the hand pointer unless we are on a link */
 div.similar_jobs .left_panel table tr { cursor: default; }
 
-#notification_box{
-    position:fixed;
-    top:70px;
-    right:10px;
+#notification_box {
+    position: fixed;
+    top: 105px;
+    left: 10px;
     z-index: 1100;
 }
 
-#notification_box div.alert{
-    margin-bottom:5px;
+#notification_box div.alert {
+    padding: 10px 16px 10px 10px;
+    margin-bottom: 5px;
 }
 
-#notification_box button.close{position:relative; top:-2px; right:-8px; color:inherit}
+#notification_box button.close {
+    position: relative;
+    top: -4px;
+    right: -8px;
+    color: inherit;
+}
 
 .add-new-filter .form-group{
     margin-bottom: 5px;
--- a/webapp/app/plugins/controller.js
+++ b/webapp/app/plugins/controller.js
@@ -254,26 +254,33 @@ treeherder.controller('PluginCtrl', [
 
         $scope.retriggerJob = function() {
             if ($scope.user.loggedin) {
-                // The logic here is somewhat complicated because we need to support
-                // two use cases the first is the case where we notify a system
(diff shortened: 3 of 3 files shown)


webapp/app/js/services/main.js
```diff
@@ -184,12 +184,24 @@ treeherder.factory('thNotify', [
...
                 severity: severity,
                 sticky: sticky
             });
-            if(!sticky){
+
+            if (!sticky) {
+                if (thNotify.notifications.length > maxNsNotifications) {
+                    /* Replicate shift code to avoid pruning during rapid retriggers
+                     * otheriwse we could use $timeout(thNotify.shift) here */
+                    for(var i=0;i<thNotify.notifications.length; i++) {
```
camd: I wonder if reusing the ``shift`` code didn't work because of that ``return`` there.  If you just call ``shift`` it will return from shift, but not from this code.  which would mean it would still call the ``$timeout`` on line 205.  Maybe that was the problem?

maurodoglio: Why don't you call thNotify.shift directly instead of copy/paste?

tojonmz: @maurodoglio ya, I had included this explanation in my initial PR description:
*"with many rapid notifications, it appears that $timeout(thNotify.shift) will start to do weird things, pruning the queue where the 4000ms timeout should not have occurred. So I chose to duplicate the shift code, which appears to work perfectly."*

But ya, I will have a look at that `return` that @camd mentions. Thanks!

maurodoglio: I'm probably missing a piece here... can you please help me to understand why executing 
thNotify.shift() is different from executing its body? Is that what's happening?

tojonmz: During testing with thNotify.shift(), the notifications will begin to not populate all 5 potential notification spots on screen (you will see 4, perhaps 3 notifications over a period of 15-30secs of flooding the retrigger 'queue'). If instead you execute the duplicated body code instead, it's fine and all 5 notification spots continue to cascade/populate on screen. I will try though what @camd suggests.

tojonmz: Yeah, it looks like with that extra `return` I can achieve the correct behavior I had with duplicated shift code. Excellent, thank you sirs! I will switch it and make that change.

tojonmz:
We've had Will, Ed and Ryan check out the interaction today on github pages, and it seems we are good to merge, so I will do that shortly.

PR #485, opened 2015-04-23
Bug 1142680 - perfherder compare subtest view

jmaher:
initial work with some refactoring to get subtests viewed from the compare view on perfherder.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/485)

### DIFF
--- a/webapp/app/js/perf.js
+++ b/webapp/app/js/perf.js
@@ -28,3 +28,26 @@ perf.factory('getSeriesSummary', [ function() {
              options: options, subtestSignatures: subtestSignatures };
   };
 }]);
+
+perf.factory('math', [ function() {
+  return {
+    /**
+     * Compute the standard deviation for an array of values.
+     *
+     * @param values
+     *        An array of numbers.
+     * @param avg
+     *        Average of the values.
+     * @return a number (the standard deviation)
+     */
+    stddev: function(values, avg) {
+      if (values.length <= 1) {
+        return 0;
+      }
+
+      return Math.sqrt(
+        values.map(function (v) { return Math.pow(v - avg, 2); })
+          .reduce(function (a, b) { return a + b; }) / (values.length - 1));
+    }
+  };
+}]);
--- a/webapp/app/js/perf.js
+++ b/webapp/app/js/perf.js
@@ -6,28 +6,241 @@
 
 var perf = angular.module("perf", ['ui.router', 'ui.bootstrap', 'treeherder']);
 
-perf.factory('getSeriesSummary', [ function() {
-  return function(signature, signatureProps, optionCollectionMap) {
-    var platform = signatureProps.machine_platform + " " +
-      signatureProps.machine_architecture;
-    var e10s = (signatureProps.job_group_symbol === "T-e10s");
-    var testName = signatureProps.test;
-    var subtestSignatures;
-    if (testName === undefined) {
-      testName = "summary";
-      subtestSignatures = signatureProps.subtest_signatures;
-    }
-    var name = signatureProps.suite + " " + testName;
-    var options = [ optionCollectionMap[signatureProps.option_collection_hash] ];
-    if (e10s) {
-      options.push("e10s");
-    }
-    name = name + " " + options.join(" ");
+perf.factory('PhSeries', ['$http', 'thServiceDomain', function($http, thServiceDomain) {
 
-    return { name: name, signature: signature, platform: platform,
-             options: options, subtestSignatures: subtestSignatures };
+  var _getSeriesSummary = function(signature, signatureProps, optionCollectionMap) {
+      var platform = signatureProps.machine_platform + " " +
+        signatureProps.machine_architecture;
+      var e10s = (signatureProps.job_group_symbol === "T-e10s");
+      var testName = signatureProps.test;
+      var subtestSignatures;
+      if (testName === undefined) {
+        testName = "summary";
+        subtestSignatures = signatureProps.subtest_signatures;
+      }
+      var name = signatureProps.suite + " " + testName;
+      var options = [ optionCollectionMap[signatureProps.option_collection_hash] ];
+      if (e10s) {
+        options.push("e10s");
+      }
+      name = name + " " + options.join(" ");
+
+      return { name: name, signature: signature, platform: platform,
+               options: options, subtestSignatures: subtestSignatures };
+  };
+
+  var _getAllSeries = function(projectName, timeRange, optionMap) {
+    var signatureURL = thServiceDomain + '/api/project/' + projectName + 
+      '/performance-data/0/get_performance_series_summary/?interval=' +
+      timeRange;
+
+    return $http.get(signatureURL).then(function(response) {
+      var seriesList = [];
+      var platformList = [];
+      var testList = [];
+
+      Object.keys(response.data).forEach(function(signature) {
+        var seriesSummary = _getSeriesSummary(signature,
+                                             response.data[signature],
+                                             optionMap);
+
+        seriesList.push(seriesSummary);
+
+        // add test/platform to lists if not yet present
+        if (!_.contains(platformList, seriesSummary.platform)) {
+          platformList.push(seriesSummary.platform);
+        }
+        if (!_.contains(testList, seriesSummary.name)) {
+          testList.push(seriesSummary.name);
+        }
+      });
+
+      return {
+        seriesList: seriesList,
+        platformList: platformList,
+        testList: testList
+      };
+    });
+  };
+
+  return {
+    getSeriesSummary: function(signature, signatureProps, optionCollectionMap) {
+      return _getSeriesSummary(signature, signatureProps, optionCollectionMap);
+    },
+
+    getSubtestSummaries: function(projectName, timeRange, optionMap, targetSignature) {
+      return _getAllSeries(projectName, timeRange, optionMap).then(function(lists) {
+        var seriesList = [];
+        var platformList = [];
+        var subtestSignatures = [];
+        var suiteName = "";
+
+        //Given a signature, find the series and get subtest signatures
+        var seriesSummary = _.find(lists.seriesList,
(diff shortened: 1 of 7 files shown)


webapp/app/js/perf.js
```diff
@@ -6,6 +6,19 @@
 
 var perf = angular.module("perf", ['ui.router', 'ui.bootstrap', 'treeherder']);
 
+perf.factory('isReverseTest', [ function() {
```
wlach: We should really be picking this up in the signature. Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1158233 for this

webapp/app/js/perf.js
```diff
@@ -28,3 +41,207 @@ perf.factory('getSeriesSummary', [ function() {
              options: options, subtestSignatures: subtestSignatures };
   };
 }]);
+
+
+perf.factory('getCounterMap', [ 'isReverseTest',
+  function(isReverseTest) {
+    return function(testName, originalData, newData) {
+      var cmap = {'originalGeoMean': NaN, 'originalRuns': 0, 'originalStddev': NaN,
```
wlach: I know this is just copypasted code, but could you take out the quotes. i.e. { 'originalGeoMean': ... } -> { orginalGeoMean: ... }

webapp/app/js/perf.js
```diff
@@ -28,3 +41,207 @@ perf.factory('getSeriesSummary', [ function() {
...
+    var signatureURL = thServiceDomain + '/api/project/' + projectName + 
+      '/performance-data/0/get_performance_series_summary/?interval=' +
+      timeRange;
+    var e10s = false;
+    var pgo = false;
+    var summaryOnly = false;
+    var targetSignature = '';
+
+    if (userOptions.e10s != undefined)
+      e10s = userOptions.e10s;
```
wlach: Both 'undefined' and 'false' evaluate to false, so I don't think the if clauses are necessary here.

webapp/app/js/perf.js
```diff
@@ -28,3 +41,207 @@ perf.factory('getSeriesSummary', [ function() {
...
+    var avg = total / geomeans.length;
+    var sigma = math.stddev(geomeans, avg);
+    return {geomean: avg.toFixed(2), stddev: sigma.toFixed(2), runs: geomeans.length};
+  }
+}]);
+
+
+perf.factory('getSeriesLists', [ '$http', 'getSeriesSummary', 'thServiceDomain',
+  function($http, getSeriesSummary, thServiceDomain) {
+  return function(projectName, timeRange, optionMap, userOptions) {
```
wlach: I have to admit I find this method rather confusing. I think it would be easier if we broke out separate methods for: 

(1) getting all the summary signatures
(2) getting the subtest signatures for a particular summary signature

It might make sense to make both of these methods part of a higher level PhSeries object (see e.g. https://github.com/mozilla/treeherder-ui/blob/master/webapp/app/js/models/resultset.js#L7 for an example of this pattern), so you could reuse some code between them (as well as have less to declare as dependencies in the controllers).

webapp/app/js/perf.js
```diff
@@ -28,3 +41,207 @@ perf.factory('getSeriesSummary', [ function() {
...
+        seriesList: seriesList,
+        platformList: platformList,
+        testList: testList
+      };
+    });
+  };
+}]);
+
+
+perf.factory('getResultsMap', [ '$q', '$http', 'thServiceDomain', 'calculateStats',
```
wlach: This method could probably also be included in the PhSeries object I propose above.

webapp/app/partials/perf/comparectrl.html
```diff
@@ -19,10 +19,12 @@
       <h4>Compare all data for revision {{newProject}}: {{newRevision}} to {{originalProject}}: {{originalRevision}}</h4>
       <p class="help-block">Hover over each entry for more options.</p>
       <table class="table">
-        <tbody>
-          <tr ng-class="{'subtest-header': compareResult.headerColumns==2, 'subtest-empty': compareResult.isEmpty && !compareResult.headerColumns==2, 'subtest-empty': (hideMinorChanges && compareResult.isMinor)}" ng-repeat="compareResult in compareResults">
-            <td colspan="{{compareResult.headerColumns}}">{{compareResult.name}}</td>
-            <td ng-class="{'subtest-empty': compareResult.headerColumns==2}"><a ng-href="{{compareResult.detailsLink}}">Details</a></td>
+        <tbody ng-repeat="testName in testList">
+          <tr class="subtest-header">
+            <td>{{testName.replace('summary ', '')}}</td><td>Old Geomean</td><td>Old StdDev</td><td>New Geomean</td><td>New StdDev</td><td>Delta</td><td>Delta %</td>
```
wlach: Putting a .replace() inside the template is pretty gross. Is there no way we could generate the required string in the code somehow? I guess the best thing would be to store the name stripped of keywords like "summary" and just use that here,

webapp/app/js/perf.js
```diff
@@ -28,3 +41,207 @@ perf.factory('getSeriesSummary', [ function() {
...
+        } else if (cmap.deltaPercentage < -2.0) {
+          cmap.isMinor = false;
+          isReverseTest(testName) ? cmap.isRegression = true : cmap.isImprovement = true;
+        }
+      }
+      return cmap;
+    }
+}]);
+
+perf.factory('calculateStats', [ 'math', function(math) {
```
wlach: I find the name calculateStats to be a bit too vague for my liking when we're dealing with something this big (this certainly isn't the only place we're calculating stats). It appears that we're only using this method in one place (in getResultsMap). Maybe we'd be better off just inlining it there? If we're going to keep it, I'd probably rename it to something like getDataPointStats.

webapp/app/js/perf.js
```diff
@@ -28,3 +41,207 @@ perf.factory('getSeriesSummary', [ function() {
...
+    _.where(perfData, { result_set_id: resultSetID }).forEach(function(pdata) {
+      //summary series have geomean, individual pages have mean
+      if (pdata.geomean === undefined) {
+        geomeans.push(pdata.mean);
+      } else {
+        geomeans.push(pdata.geomean);
+      }
+    });
+
+    geomeans.forEach(function(mean) { total += mean; });
```
wlach: You could express this slightly more concisely as:

    var total = _.reduce(means, function(mean, total) { return total + mean; })

webapp/app/js/perf.js
```diff
@@ -28,3 +41,207 @@ perf.factory('getSeriesSummary', [ function() {
...
+          isReverseTest(testName) ? cmap.isRegression = true : cmap.isImprovement = true;
+        }
+      }
+      return cmap;
+    }
+}]);
+
+perf.factory('calculateStats', [ 'math', function(math) {
+  return function(perfData, resultSetID) {
+    var geomeans = [];
```
wlach: geomeans -> means, since we're not always dealing with geometric means

webapp/app/js/perf.js
```diff
@@ -28,3 +41,207 @@ perf.factory('getSeriesSummary', [ function() {
...
+      } else {
+        geomeans.push(pdata.geomean);
+      }
+    });
+
+    geomeans.forEach(function(mean) { total += mean; });
+
+    var avg = total / geomeans.length;
+    var sigma = math.stddev(geomeans, avg);
+    return {geomean: avg.toFixed(2), stddev: sigma.toFixed(2), runs: geomeans.length};
```
wlach: I know this was probably the case before, but it's probably not the best idea to run .toFixed() on a value until we're actually displaying it. For that, it looks like we could write an angular filter and then put that in our template:

See e.g. https://github.com/mozilla/treeherder-ui/blob/master/webapp/app/js/filters.js#L40 and https://github.com/mozilla/treeherder-ui/blob/master/webapp/app/partials/main/thWatchedRepoInfoDropDown.html#L3

webapp/app/js/perf.js
```diff
@@ -6,25 +6,273 @@
...
-    var subtestSignatures;
-    if (testName === undefined) {
-      testName = "summary";
-      subtestSignatures = signatureProps.subtest_signatures;
-    }
-    var name = signatureProps.suite + " " + testName;
-    var options = [ optionCollectionMap[signatureProps.option_collection_hash] ];
-    if (e10s) {
-      options.push("e10s");
+perf.factory('phSeries', ['$http', 'thServiceDomain', function($http, thServiceDomain) {
```
wlach: For whatever reason, the convention seems to be to capitalize factories. I.e. phSeries should be PhSeries.

jmaher:
thanks for the awesome feedback, I have updated the commit with --amend and all items should be addressed.  This feels a lot cleaner now- not perfect, but many workable!

webapp/app/js/perf.js
```diff
@@ -6,25 +6,273 @@
...
+          isReverseTest(testName) ? cmap.isRegression = true : cmap.isImprovement = true;
+        }
+      }
+      return cmap;
     }
-    name = name + " " + options.join(" ");
+}]);
+
+
+perf.factory('getResultsMap', [ '$q', '$http', 'thServiceDomain', 'phSeries', 'math',
```
wlach: Let's fold getCounterMap, getResultsMap into a factory called PhCompare -- they're only really useful in that context.

webapp/app/js/perf.js
```diff
@@ -6,25 +6,273 @@
...
+
+      return Math.sqrt(
+        values.map(function (v) { return Math.pow(v - avg, 2); })
+          .reduce(function (a, b) { return a + b; }) / (values.length - 1));
+    }
   };
 }]);
+
+
+perf.filter('precision', function() {
```
wlach: Let's call this displayPrecision, a bit more... precise :)

jmaher:
updated PR with --amend, this should be rad!

wlach:
Merged manually.

PR #486, opened 2015-04-26
Speed improvements

fitzgen:
## Before

screen shot 2015-04-26 at 3 56 06 pm

## After

screen shot 2015-04-26 at 3 54 06 pm

That one last FPS dip is completely within angular. I suspect making it go away would involve significant changes within angular, or completely ditching angular in treeherder.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/486)

### DIFF
--- a/webapp/app/js/models/job.js
+++ b/webapp/app/js/models/job.js
@@ -66,10 +66,14 @@ treeherder.factory('ThJobModel', [
                         return new ThJobModel(job_obj);
                     });
                 }
+
                 // next_pages_jobs is wrapped in a $q.when call because it could be
                 // either a promise or a value
-                return $q.when(next_pages_jobs).then(function(maybe_job_list){
-                    return  item_list.concat(maybe_job_list);
+                return $q.when(next_pages_jobs).then(function(maybe_job_list) {
+                    for (var i = 0, n = maybe_job_list.length; i < n; i++) {
+                        item_list.push(maybe_job_list[i]);
+                    }
+                    return  item_list;
                 })
         });
     };
--- a/webapp/app/js/models/job.js
+++ b/webapp/app/js/models/job.js
@@ -57,10 +57,17 @@ treeherder.factory('ThJobModel', [
                 if(_.has(response.data, 'job_property_names')){
                     // the results came as list of fields
                     //we need to convert them to objects
-                    item_list = _.map(response.data.results, function(elem){
-                        var job_obj = _.object(response.data.job_property_names, elem);
-                        return new ThJobModel(job_obj);
-                    });
+                    item_list = [];
+                    var props = response.data.job_property_names;
+                    var propsLen = props.length;
+                    for (var i = 0, resultsLen = response.data.results.length; i < propsLen; i++) {
+                        var model = new ThJobModel();
+                        var elem = response.data.results[i];
+                        for (var j = 0; j < propsLen; j++) {
+                            model[props[j]] = elem[j];
+                        }
+                        item_list.push(model);
+                    }
                 }else{
                     item_list = _.map(response.data.results, function(job_obj){
                         return new ThJobModel(job_obj);
--- a/webapp/app/js/models/job.js
+++ b/webapp/app/js/models/job.js
@@ -10,9 +10,11 @@ treeherder.factory('ThJobModel', [
     // ThJobModel is the js counterpart of job
 
     var ThJobModel = function(data) {
-        // creates a new instance of ThJobModel
-        // using the provided properties
-        angular.extend(this, data);
+        if (typeof data === "object" && data !== null) {
+            for (var k in data) {
+                this[k] = data[k];
+            }
+        }
     };
 
     ThJobModel.prototype.get_current_eta = function(){
--- a/webapp/app/vendor/jquery-2.0.3.js
+++ b/webapp/app/vendor/jquery-2.0.3.js
@@ -4789,7 +4789,7 @@ jQuery.event = {
 		}
 
 		// Create a writable copy of the event object and normalize some properties
-		var i, prop, copy,
+		var i, copy,
 			type = event.type,
 			originalEvent = event,
 			fixHook = this.fixHooks[ type ];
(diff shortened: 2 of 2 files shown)


fitzgen:
Also, I couldn't for the life of me get the unit tests to run. This is what I get:

    $ ./scripts/test.sh 
    
    Starting Karma Server (http://karma-runner.github.io)
    -------------------------------------------------------------------
    INFO [karma]: Karma v0.12.31 server started at http://localhost:9876/
    INFO [launcher]: Starting browser Firefox
    INFO [Firefox 36.0.0 (Mac OS X 10.10)]: Connected on socket KFbF3sPG-CNXcyJF663y with id 84325436
    Firefox 36.0.0 (Mac OS X 10.10) ERROR
      TypeError: jasmine.Matchers is undefined
      at /Users/fitzgen/src/treeherder-ui/webapp/test/vendor/jasmine-jquery.js:530

fitzgen:
Hrm, seems the tests failed on travis-ci, so maybe these results are bunk (but everything seems to work fine locally for me).

If anyone has ideas on why the unit tests don't run locally for me, I'm all ears...

tojonmz:
@fitzgen those results look amazing. Is there an open bug to which I can assign this PR? We have a meta UI performance bug [1067846](https://bugzilla.mozilla.org/show_bug.cgi?id=1067846), and a depends bug [1125229](https://bugzilla.mozilla.org/show_bug.cgi?id=1125229) in case either of those are applicable.

I defer to @maurodoglio and @edmorley on the current CI failures.

maurodoglio:
@fitzgen I tried your branch locally, it's a big improvement indeed :smiley: , thank you!
I'll try to figure out why the travis run is failing.
Regarding the error you get locally, you may be using an outdated version of jasmine/jasmine-jquery. From a fresh node environment npm install && npm test is all you need to run the tests. We probably need to pin some more packages in package.json :disappointed:

webapp/app/js/models/job.js
```diff
@@ -55,19 +57,30 @@ treeherder.factory('ThJobModel', [
...
                     // the results came as list of fields
                     //we need to convert them to objects
-                    item_list = _.map(response.data.results, function(elem){
-                        var job_obj = _.object(response.data.job_property_names, elem);
-                        return new ThJobModel(job_obj);
-                    });
+                    item_list = [];
+                    var props = response.data.job_property_names;
+                    var propsLen = props.length;
+                    for (var i = 0, resultsLen = response.data.results.length; i < propsLen; i++) {
```
maurodoglio: The condition to break the loop should be i < resultsLen 
I think this is the reason why the tests are failing. You probably don't see the problem if you look at `mozilla-central`, since the number of jobs per result set is > 36. On `try` you will likely see errors in the console.

edmorley:
Awesome - thank you for digging into this!
For the jquery change - is there an upstream issue open for it? Would be awesome to get it fixed upstream if possible :-)

webapp/app/js/models/job.js
```diff
@@ -10,9 +10,11 @@ treeherder.factory('ThJobModel', [
     // ThJobModel is the js counterpart of job
 
     var ThJobModel = function(data) {
-        // creates a new instance of ThJobModel
-        // using the provided properties
-        angular.extend(this, data);
+        if (typeof data === "object" && data !== null) {
```
wlach: We need a comment here to explain why we're not using angular.extend, otherwise someone might be tempted to change it back (since the original form is more concise).

webapp/app/js/models/job.js
wlach: Before breaking this stuff out into loops (and making the code much less readable), I wonder if we should consider using lodash (https://lodash.com/) instead of underscore. It's supposed to be quite a bit faster than underscore (http://benmccormick.org/2014/11/12/underscore-vs-lodash/), and indeed uses the technique you're using here to iterate through arrays with map (and other things): https://github.com/lodash/lodash/blob/master/lodash.js#L1443

wlach:
So first off, thanks for doing this! I have some concerns about some of the implementation details, but the basic ideas seem good. Some questions:

1. How were you interacting with treeherder when you were generating the above profiles?
2. What relative speed increases did you get from each of the above commits? 

As @tojonmz mentioned, we should open a bug to associate this work with.

webapp/app/js/models/job.js
fitzgen: > The condition to break the loop should be i < resultsLen
> I think this is the reason why the tests are failing.

Yes! Thanks! All passing now.

fitzgen:
> How were you interacting with treeherder when you were generating the above profiles?

This is scrolling to the bottom, start profiling, clicking on "load 50 more", waiting for them all to load, end profiling.

> What relative speed increases did you get from each of the above commits?

Mostly from ditching underscore to (a) avoid allocations and reduce GC pressure, and (b) get better interaction with the JIT. That's the big FPS gain you can see throughout the middle of the recording. The change to push instead of concat helps ease GC pressure here as well.

We could probably ditch the `ThJobModel` constructor change and instead have a factory function something like this:

    ThJobModel.constructWithProperties = function (properties, values) {
        var model = new ThJobModel();
        ... add properties ...
        return model;
    };

That would clean up the code organization a little bit so that the properties aren't added in the callback and things are a bit better encapsulated.

The jQuery fix causes that initial FPS dip to go from ~300ms down to ~150ms. I talked to @mzgol on twitter, not sure if he filed a bug.

> npm install && npm test

Ok, that worked for me, thanks. FWIW, readthedocs suggests other stuff with `-g` and craziness.

fitzgen:
Ok, so as I feared, the fact that the tests were failing and UI responsiveness improved were related :)

We were creating a ton less models because we were only making as many models as there are properties. With the bug fixed, these changes (other than the jQuery one) don't actually help too much. Back to the drawing board!

mzgol:
@fitzgen do you happen to have a relatively small example that shows the performance penalty of copying event properties in Firefox? I'd like to dive in but I need something to start from and preferably something that is not extremely huge. :)

fitzgen:
Sorry, I don't. I suspect it isn't a bottleneck unless there is a bunch of stuff requiring reflow, so a smaller test case may be hard to create.

PR #490, opened 2015-04-27
Bug 1142680 - Write a compare-talos equivalent - optimizations and bar graphs. r=wlach

jmaher:
optimizations for:
* loading one set of data if we have the same branch
* only requesting the amount of data that is needed (not defaulting to 30 days)
* batch up network requests (20 signatures at a time) to reduce network requests.
* add a bar chart view to overview and subtests (helps point out larger offenders.

tested with:
http://localhost:8000/app/perf.html#/compare?originalProject=try&originalRevision=76b999af7321&newProject=try&newRevision=59b75c0e8e97

optimizations:
* instead of 314+ network requests, we have 13 for the above link :)

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/490)


webapp/app/js/perf.js
```diff
@@ -203,35 +217,57 @@ perf.factory('PhCompare', [ '$q', '$http', 'thServiceDomain', 'PhSeries', 'math'
...
     },
 
-    getResultsMap: function(projectName, seriesList, timeRange, resultSetId) {
+    getInterval: function(oldTimestamp, newTimestamp) {
+      var now = (new Date()).getTime() / 1000;
+      var timeRange = Math.min(oldTimestamp, newTimestamp);
+      timeRange = Math.round(now - timeRange);
+
+      //now figure out which predefined set of data we can query from
+      var intervals = [86400, 604800, 1209600, 2592000, 5184000, 7776000];
```
wlach: So this is already used here: https://github.com/mozilla/treeherder-ui/blob/master/webapp/app/js/graphs.js#L16

We should create an angular value for this so we can reuse this information here, there, and elsewhere. The keys should be the human readable name, the value the integer values we send to treeherder/perfherder (for cases we just need the values, just get the values). Some examples here: https://github.com/jmaher/treeherder-ui/blob/subtest/webapp/app/js/values.js and

webapp/app/js/perf.js
```diff
@@ -203,35 +217,57 @@ perf.factory('PhCompare', [ '$q', '$http', 'thServiceDomain', 'PhSeries', 'math'
...
+    getResultsMap: function(projectName, seriesList, timeRange, resultSetIds) {
       var baseURL = thServiceDomain + '/api/project/' +
         projectName + '/performance-data/0/' +
         'get_performance_data/?interval_seconds=' + timeRange;
 
       var resultsMap = {};
-      return $q.all(seriesList.map(function(series) {
-        return $http.get(baseURL + "&signatures=" + series.signature).then(
+      var chunks = _.chunk(seriesList, 20);
+      return $q.all(chunks.map(function(seriesChunk) {
```
wlach: May as well make this all one expression: return $q.all(_chunk(seriesList, 20).map(...)

jmaher:
we can close this PR as #491 is used with a real branch, etc.

PR #495, opened 2015-04-29
Bug 1142680 - Write a compare-talos equivalent - items 1,2,7,8

jmaher:
[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/495)

### DIFF
--- a/webapp/app/js/compareperf.js
+++ b/webapp/app/js/compareperf.js
@@ -59,7 +59,7 @@ perf.controller('CompareResultsCtrl', [
           var originalResultsMap = resultMaps[$scope.originalResultSetID];
           var newResultsMap = resultMaps[$scope.newResultSetID];
 
-          // Optimization - collect all data in a single pass
+          // Optimization - we collected all data in a single pass
           if (newResultsMap) {
             $scope.dataLoading = false;
             displayResults(originalResultsMap, newResultsMap);
@@ -80,7 +80,8 @@ perf.controller('CompareResultsCtrl', [
                                    newSeriesData.seriesList,
                                    timeRange,
                                    [$scope.newResultSetID]);
-            }).then(function(newResultsMap) {
+            }).then(function(resultMaps) {
+              var newResultsMap = resultMaps[$scope.newResultSetID];
               $scope.dataLoading = false;
               displayResults(originalResultsMap, newResultsMap);
             });
@@ -116,6 +117,7 @@ perf.controller('CompareResultsCtrl', [
 
           cmap.detailsLink = detailsLink;
           cmap.name = platform;
+          cmap.hideMinorChanges = $scope.hideMinorChanges;
           $scope.compareResults[testName].push(cmap);
         });
       });
@@ -143,8 +145,9 @@ perf.controller('CompareResultsCtrl', [
       });
     }
 
-    var optionCollectionMap = {};
     $scope.dataLoading = true;
+    var optionCollectionMap = {};
+    $scope.getCompareClasses = PhCompare.getCompareClasses;
 
     $http.get(thServiceDomain + '/api/optioncollectionhash').then(
       function(response) {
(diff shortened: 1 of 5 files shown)


webapp/app/js/compareperf.js
```diff
@@ -143,8 +144,17 @@ perf.controller('CompareResultsCtrl', [
       });
     }
 
-    var optionCollectionMap = {};
     $scope.dataLoading = true;
+    $scope.getDeltaClass = function(cr) { return cr.className; };
```
wlach: I'm not sure about putting CSS logic inside the controller, but it doesn't look like angular provides better options (http://stackoverflow.com/questions/7792652/what-is-the-best-way-to-conditionally-apply-a-class). Could you at least consolidate these into one method like $scope.getCompareResultClasses() ? Just append the applicable classes into a list, then join them into a string and return it.

webapp/app/js/compareperf.js
```diff
@@ -255,6 +278,16 @@ perf.controller('CompareSubtestResultsCtrl', [
       });
     }
 
+    $scope.dataLoading = true;
+    $scope.getDeltaClass = function(cr) { return cr.className; };
```
wlach: Code duplication with the above. Maybe we should put the function I suggested above in the compare factory and then make the $scope method a simple wrapper.

webapp/app/js/compareperf.js
```diff
@@ -311,14 +345,24 @@ perf.controller('CompareSubtestResultsCtrl', [
...
+                      }
+                    });
+                  }
+                  if (newSeriesMap) {
+                    Object.keys(newSeriesMap).forEach(function(series) {
+                      if (!_.contains($scope.pageList, newSeriesMap[series].name)) {
+                        $scope.pageList.push(newSeriesMap[series].name);
+                      }
+                    });
+                  }
```
wlach: You could put this logic in a forEach that iterates over originalSeriesMap and newSeriesMap

webapp/app/js/compareperf.js
```diff
@@ -326,19 +370,29 @@ perf.controller('CompareSubtestResultsCtrl', [
                   PhSeries.getSubtestSummaries($scope.newProject,
                                 timeRange,
                                 optionCollectionMap,
-                                $scope.originalSignature).then(
+                                $scope.newSignature).then(
                     function (newSeriesData) {
+                      $scope.platformList = _.union($scope.platformList,
+                                                    newSeriesData.platformList).sort();
+                      $scope.testList = _.union($scope.testList,
+                                                    newSeriesData.testList).sort();
```
wlach: Nit: indentation

webapp/app/js/perf.js
```diff
@@ -297,6 +334,12 @@ perf.factory('math', [ function() {
       return Math.sqrt(
         values.map(function (v) { return Math.pow(v - avg, 2); })
           .reduce(function (a, b) { return a + b; }) / (values.length - 1));
+    },
+
+    trimFloat: function(number) {
+      if (number === undefined)
+        return "notReady";
```
wlach: This doesn't seem like a great return value. I think we should just return undefined here, and maybe log an error.

webapp/app/js/compareperf.js
```diff
@@ -326,19 +370,29 @@ perf.controller('CompareSubtestResultsCtrl', [
...
                                            timeRange,
                                            [$scope.newResultSetID]);
                   }).then(function(newSeriesMaps) {
                     var newSeriesMap = newSeriesMaps[$scope.newResultSetID];
-                    Object.keys(newSeriesMap).forEach(function(series) {
-                      if (!_.contains($scope.pageList, newSeriesMap[series].name)) {
-                        $scope.pageList.push(newSeriesMap[series].name);
-                      }
-                    });
+                    if (newSeriesMap) {
```
wlach: Under what conditions would newSeriesMap not be defined? It kind of looks like we might be papering over something that shouldn't happen?

jmaher: updated the PR to address feedback!

On Wed, Apr 29, 2015 at 12:07 PM, William Lachance <notifications@github.com

PR #496, opened 2015-04-29
Bug 1057260 - Display full Job name in retrigger notification

tojonmz:
This fixes Bugzilla bug [1057260](https://bugzilla.mozilla.org/show_bug.cgi?id=1057260).

Per the bug request, this provides the full job name, in the retrigger success notification.

Here's current production:

retriggerjobcurrent

Here's proposed:

retriggerjobfullname

The implication of this change is at narrow browser widths, retrigger notifications may end up again partly obscuring parts of the job table.

Everything seems to print fine on both browsers, as one would expect. 

Tested on OSX 10.10.3:
FF Release **37.0.2**
Chrome Latest Release **42.0.2311.90** (64-bit)

Adding @wlach for review.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/496)

### DIFF
--- a/webapp/app/plugins/controller.js
+++ b/webapp/app/plugins/controller.js
@@ -270,7 +270,7 @@ treeherder.controller('PluginCtrl', [
                             return thBuildApi.retriggerJob($scope.repoName, requestId);
                         }
                     }).then(function() {
-                        thNotify.send("Retriggered job (" + $scope.job.job_type_symbol + ")",
+                        thNotify.send("Retriggered job: " + $scope.jobSearchStr,
                                       'success');
                     }).catch(function(e) {
                         // Generic error eg. the user doesn't have LDAP access


webapp/app/plugins/controller.js
```diff
@@ -270,7 +270,7 @@ treeherder.controller('PluginCtrl', [
                             return thBuildApi.retriggerJob($scope.repoName, requestId);
                         }
                     }).then(function() {
-                        thNotify.send("Retriggered job (" + $scope.job.job_type_symbol + ")",
+                        thNotify.send("Retriggered job (" + $scope.jobSearchStr + ")",
```
wlach: Let's remove the parantheses, they look a bit weird.

tojonmz: Agreed, it looks strange with the new(?) (job_type) in paren being appended to the job name in .jobSearchStr. I will remove the outer paren. I may also try "Retriggered job: " ie. with a colon, to separate the descriptor from the job name.

tojonmz:
We've iterated and will use "Retriggered job: " with no parentheses. I will merge shortly.

jobsearcstrexample

PR #497, opened 2015-04-30
Bug 1159664 - Remove superfluous format param from resultset API calls

edmorley:
Since it's not even checked by the API.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/497)

### DIFF
--- a/webapp/test/mock/resultset_list.json
+++ b/webapp/test/mock/resultset_list.json
@@ -3,8 +3,7 @@
     "count": 10,
     "filter_params": {
       "count": "10",
-      "full": "true",
-      "format": "json"
+      "full": "true"
     },
     "repository": "mozilla-inbound"
   },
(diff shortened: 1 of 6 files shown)


webapp/test/mock/resultset_list.json
```diff
@@ -3,8 +3,7 @@
     "count": 10,
     "filter_params": {
       "count": "10",
-      "full": "true",
-      "format": "json"
+      "full": "true"
```
camd: ``full`` defaults to true in the api, so we can probably remove it anywhere it's not set to false.

camd:
Other than the thing about ``full`` this looks great.  Thanks for the cleanup!

webapp/test/mock/resultset_list.json
edmorley: I was thinking of changing the default to false, in bug 1140877 :-)

PR #500, opened 2015-05-01
Bug 1142680 - Write a compare-talos equivalent, items 5,6,9

jmaher:
Show PGO data, after IRC discussion with avih, pgo will be shown by default with no option to ignore it.  This will not be seen on a try vs. try, but try vs. m-c will show pgo data for m-c and N/A for try.

For tests that have no data on original or new revisions, we will remove those header rows and print a list at the top of the page for tests with no data, otherwise all data will be shown below.

In addition, there are basic error messages printed out for invalid branches or revisions!  It is basic, not fancy, but it at least doesn't leave the user scratching their head.

Lastly the window.document.title is update for compare and subtest compare!

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/500)

### DIFF
--- a/webapp/app/js/compareperf.js
+++ b/webapp/app/js/compareperf.js
@@ -31,7 +31,7 @@ perf.controller('CompareResultsCtrl', [
                               thServiceDomain, $http, $q,
                               $timeout, PhSeries, math,
                               isReverseTest, PhCompare) {
-    function displayComparision() {
+    function displayComparison() {
       $scope.testList = [];
       $scope.platformList = [];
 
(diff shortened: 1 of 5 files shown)


webapp/app/js/compareperf.js
```diff
@@ -118,9 +115,20 @@ perf.controller('CompareResultsCtrl', [
...
+            $scope.compareResults[testName] = [];
           $scope.compareResults[testName].push(cmap);
         });
       });
+
+      // Remove the tests with no data, report them as well; not needed for subtests
+      var noresults = [];
+      _.difference($scope.testList, Object.keys($scope.compareResults)).forEach(function(name) {
+        noresults.push(' ' + name.replace(' summary', ''));
+      });
```
wlach: You could do this all in one shot with map, instead of iterating with foreach

$scope.testNoResults = _.difference($scope.testList, Object.keys($scope.compareResults)).map(function(name) { return ' ' + name.replace(' summary', '') });

webapp/app/js/compareperf.js
```diff
@@ -220,6 +230,10 @@ perf.controller('CompareSubtestResultsCtrl', [
       $scope.compareResults = {};
       $scope.titles = {};
 
+      var subtestTitle = $scope.platformList[0].split(' ')[0];
+      subtestTitle += " " + $scope.testList[0].split(' ')[0];
+      window.document.title = subtestTitle + " Perfherder Compare Subtests";
```
wlach: I'd prefer something like: "Perfherder Compare Subtests: " + subtestTitle

PR #504, opened 2015-05-01
 	Bug 1134842 - Link to github revision/pushlog info correctly

wlach:
[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/504)

### DIFF
--- a/webapp/app/js/models/repository.js
+++ b/webapp/app/js/models/repository.js
@@ -95,8 +95,39 @@ treeherder.factory('ThRepositoryModel', [
 
             // return the promise of getting the repos
             return get_list().
-                success(function (data) {
-                    $rootScope.repos = data;
+                success(function(data) {
+                    // FIXME: only supporting github + hg for now for pushlog
+                    // + revision info (we also assume dvcs_type git==github)
+                    function Repo(props) {
+                        _.assign(this, props);
+                        if (this.dvcs_type == 'git') {
+                            // FIXME: assuming master branch, which may not
+                            // always be right -- unfortunately fixing this
+                            // requires backend changes as we're not storing
+                            // such info explicitly right now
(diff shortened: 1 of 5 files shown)


webapp/app/js/models/repository.js
```diff
@@ -95,8 +95,35 @@ treeherder.factory('ThRepositoryModel', [
...
             // return the promise of getting the repos
             return get_list().
-                success(function (data) {
-                    $rootScope.repos = data;
+                success(function(data) {
+                    // FIXME: only supporting github + hg for now for pushlog
+                    // + revision info
+                    function Repo(props) {
+                        _.assign(this, props);
+                        if (this.url.indexOf('github') >= 0) {
```
edmorley: I have a slight preference for using the dvcs_type field and hinging this all off of 'hg' vs 'git' rather than searching for substrings, but up to you.

edmorley:
Thank you for doing this :-)

There also need to be changes here:
https://github.com/mozilla/treeherder-ui/blob/8c9fa129cf3f80279a6fb05aea0477f18bb13c53/webapp/app/index.html#L188
and:
https://github.com/mozilla/treeherder-ui/blob/8c9fa129cf3f80279a6fb05aea0477f18bb13c53/webapp/app/partials/perf/perfctrl.html#L69
(the latter having only just been added)

webapp/app/js/models/repository.js
wlach: Yeah, that's probably slightly better. I'll change it.

wlach:
Thanks for catching those, will fix before pushing!

wlach:
Ok, so fixing the url in perf is more difficult than it should be because it doesn't currently use ThRepositoryModel. I have plans to do that next week, let's take care of that then.

PR #506, opened 2015-05-04
Bug 1161264 - when clicking a compare view of an individual subtest we fail

jmaher:
[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/506)

### DIFF
--- a/webapp/app/css/perf.css
+++ b/webapp/app/css/perf.css
@@ -182,6 +182,10 @@ to { opacity: 0; }
   font-weight: bold;
 }
 
+.subtest-highlighted {
+  font-weight: bold;
+}
+
 .subtest-empty {
     display: none;
 }
--- a/webapp/app/js/perf.js
+++ b/webapp/app/js/perf.js
@@ -76,14 +76,27 @@ perf.factory('PhSeries', ['$http', 'thServiceDomain', function($http, thServiceD
         var suiteName;
 
         //Given a signature, find the series and get subtest signatures
-        var seriesSummary = _.find(lists.seriesList,
+        var series = _.find(lists.seriesList,
           function(series) {
             return series.signature == targetSignature;
           });
 
-        if (seriesSummary) {
-          subtestSignatures = seriesSummary.subtestSignatures;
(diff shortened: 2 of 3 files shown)


webapp/app/css/perf.css
```diff
@@ -182,6 +182,10 @@ to { opacity: 0; }
   font-weight: bold;
 }
 
+.subtest-bold {
```
wlach: let's call this something like subtest-highlighted

webapp/app/js/perf.js
```diff
@@ -82,6 +82,14 @@ perf.factory('PhSeries', ['$http', 'thServiceDomain', function($http, thServiceD
           });
 
         if (seriesSummary) {
+          if (!seriesSummary.subtestSignatures) {
+            seriesSummary = _.find(lists.seriesList,
```
wlach: It occurs to me that this won't work if we have multiple suites with the same subtests. Not currently the case, but it could happen if we changed a suite. So we should allow for multiple results here (and getting the union of all subtest signatures from them).

webapp/app/js/perf.js
```diff
@@ -82,6 +82,14 @@ perf.factory('PhSeries', ['$http', 'thServiceDomain', function($http, thServiceD
           });
 
         if (seriesSummary) {
```
wlach: The way this code is written, it kinda implies that we found a seriesSummary even if we didn't. Let's rename seriesSummary to just series.

webapp/app/js/perf.js
jmaher: looking at this a bit more, I think we might not need that.  A few things:
1) compare view for subtests doesn't support >1 header, it is designed for
1 header and a list of pages.
2) we only want one instance, the one from the graph that we clicked on :)

I am not sure right now how to get the specific data from the graph.  Given
the platform, signature, and revision- is it really likely that we will
have >1 signature?  Most likely not.  What if instead we searched for a
second instance and if there was one we add it to the list of errors or
skipped tests to be displayed?

PR #508, opened 2015-05-05
Bug 1158901 - Make off-screen job details discoverable with a tab

tojonmz:
This work fixes Bugzilla bug [1158901](https://bugzilla.mozilla.org/show_bug.cgi?id=1158901).

This makes the previously off-screen job detail content more discoverable by giving it its own "Job details" tab. We originally mused about 'Extra info' as the title, but I went with the former.

So we can now refer to two pieces of UI, "the job details pane" (where common/universal job data exists) and "the job details tab" (where more unique job data exists).

Notable changes are:

* relocating Buildername to the Job details tab (signed off by @KWierso and other sheriffs)
* providing a dyanmic Job details count, for discoverability if you aren't on that tab
* adding a divider between the two pieces of UI for all tabs

Here's before (resized to display what would otherwise be clipped off screen):

jobdetailscurrent

Here's proposed:

jobdetailsproposed

Everything seems fine in general testing on both browsers. I would like to do a bit more testing also before I merge.

Tested on OSX 10.10.3:
FF Release **37.0.2**
Chrome Latest Release **42.0.2311.135** (64-bit)

Adding @camd for review.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/508)

### DIFF
--- a/webapp/app/plugins/controller.js
+++ b/webapp/app/plugins/controller.js
@@ -126,11 +126,17 @@ treeherder.controller('PluginCtrl', [
                         // easier.
                         $scope.job_details = jobInfoArtifact.reduce(function(result, artifact) {
                           if (artifact.blob && Array.isArray(artifact.blob.job_details)) {
-                            result = result.concat(artifact.blob.job_details);
+                              result = result.concat(artifact.blob.job_details);
+                          }
+                          if ($scope.artifacts.buildapi) {
+                              $scope.artifacts.buildapi.blob.title = "Buildername";
+                              $scope.artifacts.buildapi.blob.value = $scope.artifacts.buildapi.blob.buildername;
+                              result = result.concat($scope.artifacts.buildapi.blob);
                           }
                           return result;
                         }, []);
                     }
+
                     // the fourth result comes from the jobLogUrl artifact
                     // exclude the json log URLs
                     $scope.job_log_urls = _.reject(results[3], {name: 'mozlog_json'});
@@ -189,11 +195,6 @@ treeherder.controller('PluginCtrl', [
                              $scope.job.build_os || undef
                 };
 
-                // if the buildername exists, add it as a field
-                if ($scope.artifacts.buildapi) {
-                    $scope.visibleFields.Buildername = $scope.artifacts.buildapi.blob.buildername;
-                }
-
                 // time fields to show in detail panel, but that should be grouped together
                 $scope.visibleTimeFields = {
                     requestTime: dateFilter($scope.job.submit_timestamp*1000,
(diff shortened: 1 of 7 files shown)


tojonmz:
I would like to handle overflow also, so I will add that in.

tojonmz:
@camd ok, so that overflow seems to be working nicely for all tab panels now. Review at your leisure :)

webapp/app/plugins/controller.js
```diff
@@ -126,11 +126,17 @@ treeherder.controller('PluginCtrl', [
                         // easier.
                         $scope.job_details = jobInfoArtifact.reduce(function(result, artifact) {
                           if (artifact.blob && Array.isArray(artifact.blob.job_details)) {
-                            result = result.concat(artifact.blob.job_details);
+                              result = result.concat(artifact.blob.job_details);
+                          }
+                          if ($scope.artifacts.buildapi) {
+                              $scope.artifacts.buildapi.blob.title = "Buildername";
+                              $scope.artifacts.buildapi.blob.value = $scope.artifacts.buildapi.blob.buildername;
+                              result = result.concat($scope.artifacts.buildapi.blob);
```
tojonmz: This is the part that needs scrutiny, conceptually. We move the buildapi blob here, and give it suitable `title` and `value` properties so it is consistent with the rest of the array objects, and so it works with the angular template which constructs the Job details content. Doing this also gives us an accurate 'count' of Job details entries in the tab.

edmorley:
I wonder if we can remove the dynamic count on the tab? The value is always going to be greater than zero for buildbot jobs, since they all have buildernames & other links - so I'm not sure that it adds much?

tojonmz:
@edmorley ya, it only seemed helpful given the default tab is 'Failure Summary' and not that tab. The feedback during job navigation seemed nice (ie. I know if there's a meaningful amount of 'stuff' there without having to click on it to find out), but I can totally remove the count if we wish.

@camd had a similar view as yours during initial testing I think. @KWierso I think liked the addition iirc. I interpreted the user initial complaint about lack of discoverability and tried to ensure we weren't just placing formerly off screen content, in a lowered tab. But maybe that is sufficient? I defer to you guys.

edmorley:
It was more that the count is going to vary between things like "6" and "8", rather than "0" and "8". As such, I'm not sure if at a glance it tells you if there is something special there you need to go and watch?

edmorley:
ie: if it was a tab called "Errors" and normally it had no count / was greyed out, except when it had errors, then a red "8" appeared - then it would make the need to read that tab quite obvious. But in this case there will always be a number (and a fairly high number), so not sure if it differentiates between "should read" and "don't bother reading".

tojonmz:
> It was more that the count is going to vary between things like "6" and "8", rather than "0" and "8". As such, I'm not sure if at a glance it tells you if there is something special there you need to go and watch?

It varies, but in my working branch Job details seem to range between 1 and.. some number usually gapped quite higher than that (eg. 1.. or 4..6..8.. etc). Unless anyone objects I'll remove it for now and we can easily add it in later if we have user feedback Job details is not discoverable enough.

PR #518, opened 2015-05-08
 	Bug 1163083 - Improve scrolling behaviour inside log viewer

wlach:
* Use flexbox for layout, which ensures that scrollbars are always inside
  viewport
* Wrap lines instead of providing horizontal scrollbars
* Trim extraneous whitespace off end of log lines before displaying

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/518)

### DIFF
--- a/webapp/app/css/logviewer.css
+++ b/webapp/app/css/logviewer.css
@@ -10,6 +10,8 @@ body {
     left: 0;
     min-width: 240px;
     padding-top: 15px;
+    display: flex;
+    flex-flow: column;
 }
 
 /* Suppress selected angular templates until they compile */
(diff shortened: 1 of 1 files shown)


camd:
I haven't actually loaded this to look at it, but I know that wrapped log lines can be misleading.  Can the wrapping lines have an indent?  So it's visually obvious?  Not sure how to do that, I must admit.

wlach:
That would likely be rather difficult to do (in fact I'm not even sure how you'd do it-- measure each line manually somehow?) and I don't think it would be worth it. As I mentioned, the default behavior of Firefox with text files is to wrap. Almost all terminal applications wrap too. I think it's ok for Treeherder to do the same. :)

Here's what the original offending log looks like with line wraps:
screen shot 2015-05-09 at 8 36 16 am

edmorley:
> Trim extraneous whitespace off end of log lines before displaying

I think we need to omit this part, otherwise we'll regress bug 1081587.
Things like stacktraces rely on the leading whitespace.

wlach:
Yeah, I realized that after writing this patch. We could modify the patch to trim just the trailing whitespace, but it's probably not worth it.

PR #519, opened 2015-05-11
Bug 1080760 - Failure Summary tab bug suggestions polling

camd:
This implements polling for the ``Bug suggestions`` artifact.  

Previously we checked for this artifact as soon as we detect that log parsing is complete.  But now that non-buildbot jobs can submit log summaries that require the generation of ``Bug suggestions`` artifacts, we need to not rely on looking for that artifact based on whether the status of the log is ``parsed``.  

Instead, the ``Failure Summary`` tab now does its own polling for the existence of the ``Bug suggestions`` artifact.

This also includes a commit with a couple of syntax fixes I found while I was at it.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/519)

### DIFF
--- a/webapp/app/js/models/job_log_url.js
+++ b/webapp/app/js/models/job_log_url.js
@@ -19,9 +19,9 @@ treeherder.factory('ThJobLogUrlModel', [
         if(!this.hasOwnProperty("id")){
             throw "The id property is required in order to parse a log";
         }
-        var uri = ThJobLogUrlModel.get_uri()+this.id+"/parse/"
+        var uri = ThJobLogUrlModel.get_uri()+this.id+"/parse/";
         return $http.post(uri);
-    }
+    };
 
     ThJobLogUrlModel.get_uri = function(){
         var url = thUrl.getProjectUrl("/job-log-url/");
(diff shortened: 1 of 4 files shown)


edmorley:
I'm probably missing something obvious, but why do we create the artefact synchronously when the jobs endpoint is used, but asynchronously when the artefact one is used? The commit message explains the what, but not the why? :-)

Could you add to the commit message something explaining what the different routes are (ie: artefact vs job endpoint) and who uses each? (ie: which does the log parser use?) It will help me review, since I don't entirely follow at the moment.

We also need tests for all of the permutations, particularly to ensure we don't do busy-work if the bug suggestions are already generated (eg from the log parser) :-)

edmorley:
Reviewed files:
- webapp/app/js/models/job_log_url.js @ r1
- webapp/app/plugins/controller.js @ r1
- webapp/app/plugins/failure_summary/controller.js @ r1
- webapp/app/plugins/failure_summary/main.html @ r1

---

<sup>**[webapp/app/plugins/failure_summary/main.html, line 74 \[r1\]](https://reviewable.io:443/reviews/mozilla/treeherder-ui/519#-JpDUtlSbHf-Jy8_9TjG)** ([raw file](https://github.com/mozilla/treeherder-ui/blob/d04b4439fd3d1b6764e37c2bc96826e79c87b9a9/webapp/app/plugins/failure_summary/main.html#L74)):</sup>
Should this be "generating error summary" instead of "generating bug suggestions"?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/treeherder-ui/519)

PR #520, opened 2015-05-12
Bug 1162706 - Add UI access between treeherder and perfherder

tojonmz:
This work fixes Bugzilla bug [1162706](https://bugzilla.mozilla.org/show_bug.cgi?id=1162706).

This converts our inert Treeherder and Perfherder 'logos' into equivalent nav menus, so you can switch back and forth between platforms, and discover Perfherder from Treeherder.

Here's current (inert logo):

current

Here's Treeherder proposed (menu posted):

treeherderproposed1a

We consciously use the same style as our other right hand nav bar buttons, to cue the user it is navigable, and for consistency. I don't see a downside to that approach.

Here's the proposed tooltip for both:

perftooltipproposed

Here's Perfherder proposed (menu posted):

perfproposed

The only workflow I discovered that doesn't work as expected, is if you use the Perfherder menu to navigate to Treeherder, and then hit 'back' on your browser, we just reload Treeherder instead of reloading Perfherder. The other direction is fine. I wonder if it related to our hash routing or my -ui local server development configuration. I will poke around a bit more on that. That being said, you can use the menu to easily navigate back.

Otherwise everything seems fine.

Tested on OSX 10.10.3:
FF Release **37.0.2**
FF Nightly **40.0a1**
Chrome Latest Release **42.0.2311.135** (64-bit)

Adding @wlach for review and @jmaher for visibility.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/520)

### DIFF
--- a/webapp/app/css/treeherder.css
+++ b/webapp/app/css/treeherder.css
@@ -69,17 +69,6 @@ input:focus::-moz-placeholder {
     margin-bottom: 0px;
 }
 
-.navbar-brand {
-    margin-bottom: 0px;
-    line-height: 2px;
-    height: 32px;
-    font-weight: normal;
-}
-
-.navbar-inverse .navbar-brand:hover {
-    color: #777;
-}
-
 .navbar .dropdown-menu {
     max-height: 600px;
     overflow: auto;
@@ -89,7 +78,13 @@ input:focus::-moz-placeholder {
 }
 
 .th-global-navbar {
+    padding-left: 0;
     border-bottom: 1px solid black;
+}
+
+#th-logo {
+    padding-left: 14px;
+    border-left: 0;
 }
 
 th-watched-repo {
--- a/webapp/app/css/perf.css
+++ b/webapp/app/css/perf.css
@@ -85,6 +85,13 @@ to { opacity: 0; }
   flex-flow: row;
 }
 
+#perf-logo {
+    padding: 15px;
+    border: none;
+    color: #555;
+    background: #f8f8f8;
+}
+
 #graph-chooser { flex: none; width: 240px; padding: 8px; }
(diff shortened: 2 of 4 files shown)


webapp/app/css/treeherder.css
```diff
@@ -69,17 +69,6 @@ input:focus::-moz-placeholder {
...
-    margin-bottom: 0px;
-    line-height: 2px;
-    height: 32px;
-    font-weight: normal;
-}
-
-.navbar-inverse .navbar-brand:hover {
-    color: #777;
-}
-
```
tojonmz: Retiring these bootstrap overrides, they are no longer needed.

webapp/app/partials/main/thPerfLogoMenu.html
```diff
@@ -0,0 +1,3 @@
+<li>
```
wlach: Let's just inline this for now instead of creating a partial.

webapp/app/partials/main/thLogoMenu.html
```diff
@@ -0,0 +1,3 @@
+<li>
+  <a href="perf.html">Perfherder</a>
+</li>
```
wlach: Let's just inline this for now instead of creating a partial.

webapp/app/css/treeherder.css
wlach: \o/

tojonmz:
I had one other tweak to make also; to get the treeherder link working correctly which I think will mimic its behavior on our instances. From (web-server.js) `treeherder.html` to `/#` (vagrant). It seems fine in vagrant, so I will merge now.

PR #522, opened 2015-05-12
Bug 1163580 - Titles in comparison and graphs view should reflect con…

kapy2010:
…tent

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/522)

### DIFF
--- a/webapp/app/js/compareperf.js
+++ b/webapp/app/js/compareperf.js
@@ -91,7 +91,9 @@ perf.controller('CompareResultsCtrl', [
     function displayResults(rawResultsMap, newRawResultsMap) {
       $scope.compareResults = {};
       $scope.titles = {};
-      window.document.title = "Perfherder Compare Revisions";
+      window.document.title = ($scope.originalRevision + " (" + $scope.originalProject.name + 
+                              ") " + "to " + $scope.newRevision + " (" + 
+                              $scope.newProject.name + ")");
 
       $scope.testList.forEach(function(testName) {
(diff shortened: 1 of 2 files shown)


webapp/app/js/graphs.js
```diff
@@ -432,6 +432,8 @@ perf.controller('GraphsCtrl', [
       }, {location: true, inherit: true,
           relative: $state.$current,
           notify: false});
+      window.document.title = ($scope.seriesList[0].name + " " + $scope.seriesList[0].platform + 
+                          " (" + $scope.seriesList[0].projectName + ")");
```
wlach: Actually one small request here. Could you add some code to check if the length of $scope.seriesList is greater than 1, and if so, append "and others" to window.document.title?

webapp/app/js/compareperf.js
```diff
@@ -91,7 +91,9 @@ perf.controller('CompareResultsCtrl', [
     function displayResults(rawResultsMap, newRawResultsMap) {
       $scope.compareResults = {};
       $scope.titles = {};
-      window.document.title = "Perfherder Compare Revisions";
+      window.document.title = ($scope.originalRevision + " (" + $scope.originalProject.name +
```
wlach: I can't remember what I said in the bug originally, but I think this would read slightly better if we said "Comparison between ... and ..." instead of "... to ..."

wlach:
Merged manually.

PR #528, opened 2015-05-15
Bug 1165441 - Tidy up Annotations panel layout

tojonmz:
This work fixes Bugzilla bug [1165441](https://bugzilla.mozilla.org/show_bug.cgi?id=1165441).

Specifically in the Annotations tab, it:

* fixes vertical alignment of the Classification and Bug container headers
* makes the "has not been classified" message consistent in size and style to the Failure Summary
* suppresses the data headers when there is no data so the panel is clean and simple
* makes the **Bugs** header the same computed size as the **Date**, **Author**, etc, headers

Here's current (red lines for alignment reference):
annotationscurrent

Here's proposed:
annotationsproposed

Here's current (no classifications):
annotationsemptycurrent
Here's proposed (no classifications):
annotationsemptyproposed

There's more I'll probably tweak later. Everything seems fine on all browsers.

Tested on OSX 10.10.3:
FF Release **38.0**
FF Nightly **40.0a1**

Adding @wlach for review.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/528)

### DIFF
--- a/webapp/app/css/treeherder.css
+++ b/webapp/app/css/treeherder.css
@@ -790,6 +790,7 @@ ul.failure-summary-list li .btn-xs {
 
 .failure-summary-line-empty {
     padding: 2px 4px 0px;
+    font-size: 12px;
     background: #ffffff;
 }
 
@@ -1082,6 +1083,11 @@ ul.failure-summary-list li .btn-xs {
 .bug-list {
     list-style: none inside none;
     padding-left: 0;
+}
+
+.bug-header {
+    margin-top: 8px;
+    margin-bottom: 8px;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


webapp/app/css/treeherder.css
```diff
@@ -790,6 +790,7 @@ ul.failure-summary-list li .btn-xs {
 
 .failure-summary-line-empty {
     padding: 2px 4px 0px;
+    font-size: 12px;
```
tojonmz: Here we bump up the size of the failure summary line to a 'standard' of 12px also, so they are consistent.

webapp/app/plugins/annotations/main.html
```diff
@@ -1,7 +1,7 @@
-<div ng-controller="AnnotationsPluginCtrl" class="row annotations-panel">
```
tojonmz: This `row` class was just applying a -ve margin of 15px, so I removed it. I don't think there are any ill effects.

tojonmz:
The treeherder and treeherder-ui repos were unified today. This -ui PR will continue as a new PR in treeherder.

PR #529, opened 2015-05-17
Bug 1165697 - Remove unused sample-data/*.json files

edmorley:
They were presumably used to provide dummy data when designing the UI, before the service was finished and up and running.

They are currently unused & likely out of date. I was going to move the directory around in bug 1056877, but let's just delete them. We can always re-add (up to date) mocked up API responses in the future if we need to for future UI tests.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/529)


camd:
Looks great to me!

---

Review status: :shipit: all files reviewed, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/treeherder-ui/529)

PR #532, opened 2015-05-18
Bug 1160613 - Use t-test as confidence level for highlight and display

avih:
The confidence level is used for deciding which changes to highlight (red/green)
and also for display at the table view.

New behavior:
- Ignore differences below 1.5% and low (<0.5) t-test values.
- Diff above 1.5%, and t-value below 1.0 -> not-sure-regression.
- Above t-value of 1.0 -> confident -> highlight red/green.
- Added the t-test value as the last column at the table ("Conf.")
- Added the actual runs values at the "Runs" tooltip.

TODO:
- Fine tune the t-test constants 0.5 and 1.0 if required.
- Possibly unify this calculation with our regression alerts thresholds.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/532)

### DIFF
--- a/webapp/app/js/perf.js
+++ b/webapp/app/js/perf.js
@@ -187,41 +187,45 @@ perf.factory('isReverseTest', [ function() {
 perf.factory('PhCompare', [ '$q', '$http', 'thServiceDomain', 'PhSeries',
              'math', 'isReverseTest', 'phTimeRanges',
   function($q, $http, thServiceDomain, PhSeries, math, isReverseTest, phTimeRanges) {
-  var getClassName = function(baselineMin, baselineMax, baselineAvg, currentAvg, test) {
-    var range = math.trimFloat((baselineMax - baselineMin) / 2);
-    if (isReverseTest(test)) {
-      if (currentAvg < baselineAvg - range) {
-        if (currentAvg < baselineMin) {
-          return "compare-regression";
-        }
-        // Still more than the min value we got out of the baseline, so we could
-        // be OK still, but there could be a regression here too.
-        return "compare-notsure";
-      }
-      else if (currentAvg > baselineMax) {
-        return "compare-improvement";
-      }
+
+  // Used for t_test: default stddev if both sets have only a single value - 15%.
+  // Should be rare case and it's unreliable, but at least have something.
+  var STDDEV_DEFAULT_FACTOR = 0.15;
+
+  var RATIO_CARE_MIN = 1.015; // We don't care about less than ~1.5% diff
+  var T_VALUE_CARE_MIN = 0.5; // Observations
+  var T_VALUE_CONFIDENT = 1; // Observations. Weirdly nice that ended up as 0.5 and 1...
+
+  function getClassName(newIsBetter, oldVal, newVal, abs_t_value) {
+    // NOTE: we care about general ratio rather than how much is new compared
+    // to old - this could end up with slightly higher or lower threshold
+    // in practice than indicated by DIFF_CARE_MIN. E.g.:
+    // - If old is 10 and new is 5, then new = old -50%
+    // - If old is 5 and new is 10, then new = old + 100%
+    // And if the threshold was 75% then one would matter and the other wouldn't.
+    // Instead, we treat both cases as 2.0 (general ratio), and both would matter
+    // if our threshold was 75% (i.e. DIFF_CARE_MIN = 1.75).
+    var ratio = newVal / oldVal;
+    if (ratio < 1) {
+      ratio = 1 / ratio; // Direction agnostic and always >= 1.
+    }
+
+    if (ratio < RATIO_CARE_MIN || abs_t_value < T_VALUE_CARE_MIN) {
       return "";
     }
 
-    // We have a 'smaller is better' test.
-    if (currentAvg > baselineAvg + range) {
-      if (currentAvg > baselineMax) {
-        return "compare-regression";
-      }
-      // Still less than the max value we got out of the baseline, so we could be
-      // OK still, but there could be a regression here.
-      return "compare-notsure";
+    if (abs_t_value < T_VALUE_CONFIDENT) {
+      // Since we (currently) have only one return value to indicate uncertainty,
+      // let's use it for regressions only. (Improvement would just not be marked).
+      return newIsBetter ? "" : "compare-notsure";
     }
-    else if (currentAvg < baselineMin) {
-      return "compare-improvement";
-    }
-    return "";
-  };
+
+    return newIsBetter ? "compare-improvement" : "compare-regression";
+  }
 
   return {
     getCompareClasses: function(cr, type) {
-      if (cr.hideMinorChanges && cr.isMinor) return 'subtest-empty';
+      if (cr.hideMinorChanges && !cr.isMeaningful) return 'subtest-empty';
       if (cr.isEmpty) return 'subtest-empty';
       if (type == 'row' && cr.highlightedTest) return 'active subtest-highlighted';
       if (type == 'row') return '';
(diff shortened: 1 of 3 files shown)


webapp/app/js/perf.js
```diff
@@ -368,34 +388,119 @@ perf.factory('PhCompare', [ '$q', '$http', 'thServiceDomain', 'PhSeries',
...
+        .reduce(function (a, b) { return a + b; }) / (values.length - 1));
+  }
+
+  // If a set has only one value, assume average-ish-plus sddev, which
+  // will manifest as smaller t-value the less items there are at the group
+  // (so quite small for 1 value). This default value is a parameter.
+  // C/T mean control/test group (in our case original/new data).
+  // Assumption: all the values are positive.
+  function t_test(valuesC, valuesT, stddev_default_factor) {
+      // We must have at least one value at each set
```
jmaher: nit: this function is indented 4 spaces whereas the rest of the code is 2 space; lets stick to 2 space.

webapp/app/js/perf.js
```diff
@@ -187,37 +187,40 @@ perf.factory('isReverseTest', [ function() {
...
-      }
-      else if (currentAvg > baselineMax) {
-        return "compare-improvement";
-      }
-      return "";
+
+  // Default stddev if we only have one value - 15%. Used at t_test.
+  var STDDEV_DEFAULT_FACTOR = 0.15;
+
+  var DIFF_CARE_MIN = 1.015; // We don't care about less than 1.5% diff
```
jmaher: the comment mentions 1.5%, 1.015 doesn't seem like the right constant here.

jmaher: oh, I see the comment for how 1.015 == 1.5% as a ratio.

webapp/app/js/perf.js
```diff
@@ -231,52 +234,71 @@ perf.factory('PhCompare', [ '$q', '$http', 'thServiceDomain', 'PhSeries',
...
       if (originalData) {
-         cmap.originalGeoMean = originalData.geomean;
-         cmap.originalRuns = originalData.runs;
-         cmap.originalStddev = originalData.stddev;
-         cmap.originalStddevPct = ((originalData.stddev / originalData.geomean) * 100);
-         cmap.originalMin = originalData.minVal;
-         cmap.originalMax = originalData.maxVal;
+        var orig = analyzeSet(originalData.values);
+        cmap.originalGeoMean = orig.geomean;
+        cmap.originalRuns = orig.runs + "  <  " + originalData.values.join("   ") + "  >";;
```
jmaher: originalRuns and newRuns will be an array of values?  I assume this is strictly for the mouseover feature

webapp/app/js/perf.js
avih: yes, 1.015 is the ratio below which we don't care. e.g. if old is 1.0 and new is 1.014, then that ratio would be below the threshold, but, which would be the same if if old was 1.014 and new was 1.0 - the ratio would still be below 1.05, and again, we don't care. It's ratio, but for close enough values to 1 it's close enough to percentage.

webapp/app/js/perf.js
avih: Not strictly an array, but rather the list of values, and yes - it's for the mouse-over feature. The .runs value is not used for anything else as far as I can tell.

webapp/app/js/perf.js
avih: Aye. Will do.

vitillo:
Review status: 0 of 3 files reviewed, 6 unresolved discussions, all commit checks successful.

---

<sup>**[webapp/app/js/perf.js, line 434 \[r1\]](https://reviewable.io:443/reviews/mozilla/treeherder-ui/532#-JpbG1F1Khkj0gsIiEGK)** ([raw file](https://github.com/mozilla/treeherder-ui/blob/1035bba2b9d526673423af01f7291f088d6e1ec8/webapp/app/js/perf.js#L434)):</sup>
The exponentiation of the arithmetic mean of the log-transformed values is usually the preferred way to compute the geometric mean as it's less likely to cause underflows or overflows. That said, given that we usually have few values and they are neither extremely small nor extremely large, this shouldn't be a concern.

---

<sup>**[webapp/app/js/perf.js, line 442 \[r1\]](https://reviewable.io:443/reviews/mozilla/treeherder-ui/532#-JpbIl2bGN724cicZFu_)** ([raw file](https://github.com/mozilla/treeherder-ui/blob/1035bba2b9d526673423af01f7291f088d6e1ec8/webapp/app/js/perf.js#L442)):</sup>
The standard deviation is based on the arithmetic mean, not the geometric mean.

---

<sup>**[webapp/app/js/perf.js, line 455 \[r1\]](https://reviewable.io:443/reviews/mozilla/treeherder-ui/532#-JpbHg35pmQKiQT97c--)** ([raw file](https://github.com/mozilla/treeherder-ui/blob/1035bba2b9d526673423af01f7291f088d6e1ec8/webapp/app/js/perf.js#L455)):</sup>
The correct way to use the t-test to compare two geometric means is to log-transform the data first and then apply the t-test.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/treeherder-ui/532)

webapp/app/js/perf.js
avih: >  The .runs value is not used for anything else as far as I can tell.

That's actually wrong - it's used few lines below to determine if a set is empty. But this is sort of a bug at the original code too, since if a set is empty we shouldn't have tried to calculate it's statistics to begin with. I'll fix both.

jmaher:
thanks for making these changes- I don't have any specific requests to add here.  Lets see if wlach sees something I had overlooked.

wlach:
Review status: 0 of 3 files reviewed, 14 unresolved discussions, all commit checks successful.

---

<sup>**[webapp/app/js/perf.js, line 279 \[r3\]](https://reviewable.io:443/reviews/mozilla/treeherder-ui/532#-JpgaqoRkC2MSq-sIHgK)** ([raw file](https://github.com/mozilla/treeherder-ui/blob/cff5f9ba14ca7872186495fe558ccae185d603d8/webapp/app/js/perf.js#L279)):</sup>
Let's use the "[", "]" character to demarcate the list, more universally understandable.

---

<sup>**[webapp/app/js/perf.js, line 426 \[r3\]](https://reviewable.io:443/reviews/mozilla/treeherder-ui/532#-Jpgc-rlOmH_n_Ibvbqk)** ([raw file](https://github.com/mozilla/treeherder-ui/blob/cff5f9ba14ca7872186495fe558ccae185d603d8/webapp/app/js/perf.js#L426)):</sup>
I don't really understand the point of this function, when are we going to run this outside of the context of perfherder?

If you must define it, let's just inline it inside isSetValid (or whatever we wind up calling that)

---

<sup>**[webapp/app/js/perf.js, line 434 \[r3\]](https://reviewable.io:443/reviews/mozilla/treeherder-ui/532#-JpgcHYfippLwzQOdMDM)** ([raw file](https://github.com/mozilla/treeherder-ui/blob/cff5f9ba14ca7872186495fe558ccae185d603d8/webapp/app/js/perf.js#L434)):</sup>
This function name is kinda vague and ambiguous IMO. Could you rename it to something like isSetValidForComparison?

---

<sup>**[webapp/app/js/perf.js, line 440 \[r3\]](https://reviewable.io:443/reviews/mozilla/treeherder-ui/532#-JpgbnSzJRx9j_ng1x0N)** ([raw file](https://github.com/mozilla/treeherder-ui/blob/cff5f9ba14ca7872186495fe558ccae185d603d8/webapp/app/js/perf.js#L440)):</sup>
Could you use _.each here, as we do elsewhere in the perfherder code?

---

<sup>**[webapp/app/js/perf.js, line 459 \[r3\]](https://reviewable.io:443/reviews/mozilla/treeherder-ui/532#-Jpgg64DacQWBT_6NmM5)** ([raw file](https://github.com/mozilla/treeherder-ui/blob/cff5f9ba14ca7872186495fe558ccae185d603d8/webapp/app/js/perf.js#L459)):</sup>
Let's simplify this:

return _.sum(values) / values.length;

---

<sup>**[webapp/app/js/perf.js, line 466 \[r3\]](https://reviewable.io:443/reviews/mozilla/treeherder-ui/532#-JpggH2PKiACRp2c2aTJ)** ([raw file](https://github.com/mozilla/treeherder-ui/blob/cff5f9ba14ca7872186495fe558ccae185d603d8/webapp/app/js/perf.js#L466)):</sup>
Let's take this out, as discussed.

---

<sup>**[webapp/app/partials/perf/comparesubtestctrl.html, line 17 \[r3\]](https://reviewable.io:443/reviews/mozilla/treeherder-ui/532#-Jpg_E6o8ggJWSh-EecQ)** ([raw file](https://github.com/mozilla/treeherder-ui/blob/cff5f9ba14ca7872186495fe558ccae185d603d8/webapp/app/partials/perf/comparesubtestctrl.html#L17)):</sup>
Let's not abbreviate "confidence". It's unnecessarily confusing for people who don't know what it means.

---

<sup>**[webapp/app/partials/perf/comparesubtestctrl.html, line 33 \[r3\]](https://reviewable.io:443/reviews/mozilla/treeherder-ui/532#-Jpge2-7gI5plvRU9h7g)** ([raw file](https://github.com/mozilla/treeherder-ui/blob/cff5f9ba14ca7872186495fe558ccae185d603d8/webapp/app/partials/perf/comparesubtestctrl.html#L33)):</sup>
It's really unclear what these numbers mean without looking into the source/math. Could we maybe append a string like "(low)" and "(high)" depending on whether we think that we are confident in the observation?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/treeherder-ui/532)

avih:
> Let's use the "[", "]" character to demarcate the list, more universally understandable.

That was my first attempt as well. I ended up with these symbols and spacing after trying out several variations and deciding that this one is the most readable. Specifically, `[` can be confusing if it's right next to `1`, and generous spacing instead of comma-space is much more readable. E.g. imagine the following with a small and proportional tooltip font:
[123.54, 115.12, 110.32]
compared to
<  123.54   115.12   110.32  >


> This function name is kinda vague and ambiguous IMO. Could you rename it to something like isSetValidForComparison?

That would be a misleading name since it's unrelated to comparison. It's validity test for math utils which work on a set of talos data, and it's used differently by different functions. E.g. stddev requires a minimum of two values, while average requires a minimum of one value, and neither of them is related to comparison.

I think the current name is OK, but if you have a better suggestion which would not be misleading, I'd be fine with that.


> Could you use _.each here, as we do elsewhere in the perfherder code?

Sure.


> Let's not abbreviate "confidence". It's unnecessarily confusing for people who don't know what it means.

I actually abbreviated it after consideration. The problem is the limited horizontal space since this table has quite a few columns. That being said, I also thought it could be cryptic, but eventually the space concerns won. How about we add a tooltip (either to the title or to each value) with proper description, such as `t-test value of the relation between the sets. Results are highlighted in red/green if the the delta% is above 1.5% and the t-test value is above 1.0, and are highlighted in orange if the t-test value is above 0.5`.


> It's really unclear what these numbers mean without looking into the source/math. Could we maybe append a string like "(low)" and "(high)" depending on whether we think that we are confident in the observation?

`compareResult.confidence` is a float and it's the actual value of the t-test. I could rename it instead to tTest if you like it more. The related boolean values are `.isImprovement`, `.isRegression` and `.isMeaningful`.

avih:
> > Could you use _.each here, as we do elsewhere in the perfherder code?
>
> Sure.

On second thought, this would be awkward here since the function aborts when it finds an invalid value, and aborting from within `_.each` would require jumping through (small) hoops. The straight up loop is more adequate here.

avih:
Rebased and squashed.

PR #534, opened 2015-05-18
Bug 1164166 - Create a shortcut to delete a classification

tojonmz:
This work fixes Bugzilla bug [1164166](https://bugzilla.mozilla.org/show_bug.cgi?id=1164166).

In it we provide the shortcut **ctrl+backspace** to delete a job classification if one exists. Generally everything seems to be working correctly in various testing I've done:

* rapidly deleting multiple classifications
* deleting immediately after saving the classification
* saving and deleting entirely by shortcuts (focus in main pane)
* saving via pinboard UI interaction (focus in job panel) and deleting via shortcut
* deleting classifications with and without related bugs
* deleting bug associations in Annotations, then deleting via shortcut

So here's what we have when we delete a classification via the shortcut on a job with a related bug:

deletesuccessshortcut

I've also provided a warning notification in case we end up with job which is not classified, but the UI wrongly claims it is, or the user simply invokes the shortcut on an unclassified job.

noclassificationtodeleteproposed

To me it makes sense we'd tell them :)

I have experienced one edge case where this improper display state occurs, but it is not consistently reproducible:

1. Manually pin, add a related bug, add a comment, manually click Save in pinboard
2. ctrl+backspace to delete the classification 1-2 sec after the green success msgs disappear
3. Observed: ~50% of the time the job asterisk doesn't disappear in the UI (requires a page reload)

I wonder if this is a timing thing related to either a) evalAsync() being used for save and delete, or b) non-Async emits being used in the pinboard service. Perhaps one or the other is yet to be digested.

Otherwise I haven't encountered any problems with any other workflows.

I also applied the same (perhaps slightly kludgy) fix to Save classification, simply ensuring the job is re-selected after Save so that the Annotations pane contents refresh without the user having to flip back and forth between tabs and manually unselect and re-select the job. This was a current complaint and bug reported by users in channel. Perhaps there is a better way.

Tested on OSX 10.10.3:
FF Release **38.0**
FF Nightly **40.0a1**

Adding @camd for review.

### DIFF
--- a/webapp/app/plugins/annotations/controller.js
+++ b/webapp/app/plugins/annotations/controller.js
@@ -15,10 +15,23 @@ treeherder.controller('AnnotationsPluginCtrl', [
 
         $log.debug("annotations plugin initialized");
 
-        $scope.$watch('classifications', function(newValue, oldValue){
-
+        $scope.$watch('classifications', function(newValue, oldValue) {
             thTabs.tabs.annotations.num_items = newValue ? $scope.classifications.length : 0;
         }, true);
+
+        $rootScope.$on(thEvents.deleteClassification, function(event) {
+            if ($scope.classifications[0]) {
+                $scope.deleteClassification($scope.classifications[0]);
+                // Delete any number of bugs if they exist
+                for (var i = 0; i < $scope.bugs.length; i++) {
+                    $scope.deleteBug($scope.bugs[i]);
+                }
+            } else {
+                thNotify.send("No classification on this job to delete", 'warning');
+            }
+            // We reselect job in place ensuring a correct state for other actions
+            $rootScope.$emit(thEvents.selectJob, $rootScope.selectedJob);
+        });
 
         $scope.deleteClassification = function(classification) {
 
--- a/webapp/app/plugins/pinboard.js
+++ b/webapp/app/plugins/pinboard.js
@@ -78,6 +78,8 @@ treeherder.controller('PinboardCtrl', [
             } else {
                 thNotify.send("Must be logged in to save job classifications", "danger");
             }
+            // We reselect job in place ensuring a correct state for other actions
+            $rootScope.$emit(thEvents.selectJob, $rootScope.selectedJob);
         };
 
         $scope.saveClassificationOnly = function() {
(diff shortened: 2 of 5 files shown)


webapp/app/plugins/annotations/controller.js
```diff
@@ -15,11 +15,24 @@ treeherder.controller('AnnotationsPluginCtrl', [
 
         $log.debug("annotations plugin initialized");
 
-        $scope.$watch('classifications', function(newValue, oldValue){
-
+        $scope.$watch('classifications', function(newValue, oldValue) {
```
tojonmz: Just removing an unrelated empty line here.

webapp/app/plugins/pinboard.js
```diff
@@ -78,6 +78,8 @@ treeherder.controller('PinboardCtrl', [
             } else {
                 thNotify.send("Must be logged in to save job classifications", "danger");
             }
+            // We reselect job in place ensuring a correct state for other actions
+            $rootScope.$emit(thEvents.selectJob, $rootScope.selectedJob);
```
tojonmz: This allows the Annotations tab contents to draw correctly after a classification save.

camd:
This all looks good to me.  Though I'm not quite sure what the deal is with the asterisk only being removed 1/2 the time.  That seems odd.  I think when we add the classification, we manually set the job object's ``failure_classification_id`` value to the new one.  I wonder if you might need to ensure that we're actually setting it back to 1?

If you feel stuck on that, maybe ping me in channel and we can take a look together.

tojonmz:
> Though I'm not quite sure what the deal is with the asterisk only being removed 1/2 the time.

Actually it's just ~1/2 the time that I do that fairly edge case workflow. My best guess real world perhaps < 0.5% of the time, a user might experience it (the UI still claiming the job is classified when it isn't, and you need to simply reload the page). We can check together tomorrow though in case there's something obvious to be fixed there.

tojonmz:
So the treeherder and treeherder-ui repos were unified today. This means that this -ui PR will continue as a new PR in treeherder.

PR #535, opened 2015-05-18
Bug 1166028 - Move Job Details tab to be first in the tab sequence

tojonmz:
This fixes Bugzilla bug [1166028](https://bugzilla.mozilla.org/show_bug.cgi?id=1166028).

This re-orders the new Job Details tab to be first in the tab sequence in the details panel. We preserve existing rules for Talos display.

Talos pass (unchanged, defaults to Talos tab):

passedtalosjob

Talos failure (unchanged, defaults to Failure summary):

failedtalosjob

We add new rule for all other jobs: *any other* job status type other than success lands us on the Failure Summary tab. A successful job now defaults us on the Job Details tab, since there is no 'Failure summary' failure, log parsing status, or raw log availability message to view.

All other jobs, success (changed, defaults to Job details):
passednontalos

All other jobs, eg. failure (unchanged, defaults to Failure summary);

nonpassednontalos

All other jobs, eg. pending/running (unchanged, defaults to Failure summary);

runningpendingnontalos

Everything seems fine on both browsers.

Tested on OSX 10.10.3:
FF Nightly **40.0a1**

Adding @camd for review and @KWierso and @rvandermeulen for visibility.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/treeherder-ui/535)

### DIFF
--- a/webapp/app/plugins/controller.js
+++ b/webapp/app/plugins/controller.js
@@ -91,17 +91,26 @@ treeherder.controller('PluginCtrl', [
                     $scope.eta = $scope.job.get_current_eta();
                     $scope.eta_abs = Math.abs($scope.job.get_current_eta());
                     $scope.typical_eta = $scope.job.get_typical_eta();
-                    // this is a bit hacky but for now talos is the only exception we have
-                    if($scope.job.job_group_name.indexOf('Talos') !== -1){
+
+                    // we handle which tab gets presented in the job details panel
+                    // and a special set of rules for talos
+                    if ($scope.job.job_group_name.indexOf('Talos') !== -1) {
                         $scope.tabService.tabs.talos.enabled = true;
-                        if(thResultStatus($scope.job) === 'success'){
+                        if (thResultStatus($scope.job) === 'success') {
                             $scope.tabService.selectedTab = 'talos';
-                        }else{
+                        } else {
                             $scope.tabService.selectedTab = 'failureSummary';
                         }
-                    }else{
+                    } else {
+                        // tab presentation for any other (non-talos) job
                         $scope.tabService.tabs.talos.enabled = false;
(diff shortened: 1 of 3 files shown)


webapp/app/plugins/tabs.js
```diff
@@ -34,11 +34,11 @@ treeherder.factory('thTabs', [
                     enabled: false
                 }
             },
-            "selectedTab": "failureSummary",
+            "selectedTab": "jobDetails",
             "showTab" : function(tab, contentId){
                 thTabs.selectedTab = tab;
                 if(!thTabs.tabs[thTabs.selectedTab].enabled){
-                    thTabs.selectedTab = 'failureSummary';
+                    thTabs.selectedTab = 'jobDetails';
```
tojonmz: This and line 37 are up for debate, they seem to be fine to switched to be the default (doesn't seem to produce any terrible annoying refresh-switch when loading a failed job which then loads Failure summary), but I defer to @camd.

webapp/app/plugins/controller.js
```diff
@@ -91,17 +91,26 @@ treeherder.controller('PluginCtrl', [
...
                         }
-                    }else{
+                    } else {
+                        // tab presentation for any other (non-talos) job
                         $scope.tabService.tabs.talos.enabled = false;
+                        if (thResultStatus($scope.job) === 'success') {
+                            $scope.tabService.selectedTab = 'jobDetails';
+                        } else {
+                            $scope.tabService.selectedTab = 'failureSummary';
+                        }
```
tojonmz: Actually I think I could avoid duplicating here and rearrange this all to be a bit more economical.

tojonmz: I had a look, we have to repeat code the other way also. It's probably more clear this way if we have to add more conditions for other job status's later, so I will leave it as is.

